PR #12, opened 2019-07-11
Design/minor fixes

JosefBud:
https://github.com/gusta-project/frontend/commit/7efa162bb8324e638aa334b68e9030a4e3721089 React was throwing an error in the Chrome development panel because the list items weren't keyed. Even though this is a placeholder list, I gave them keys to get rid of the error.

Also removed the hover effect for the table cells.

https://github.com/gusta-project/frontend/commit/5917e928e2790dd7a378cc23ac467a41c8f6ad3e Small change to style.scss so that the striped tables fit with the color scheme. The default gray cells are now `$lightest-blue` instead.

https://github.com/gusta-project/frontend/commit/666cd8537b3e8315fd6db8f9ba745d09431990d5 React was throwing an error in the Chrome development panel  because it didn't like me using the standard CSS `"font-size"` object key for inline styling, so I changed it to `fontSize`. It made no visual difference, it just removed the error.

https://github.com/gusta-project/frontend/commit/846ee8f1ff66c1d2d4cd2ceb5e65d725501cb7e1 Updated test snapshot for the small `fontSize` change in 666cd85

### DIFF
--- a/src/pages/user/Recipes.js
+++ b/src/pages/user/Recipes.js
@@ -10,7 +10,9 @@ export default class UserRecipes extends Component {
     for (let i = 0; i < 20; i++) {
       listItems.push(
         <tr key={i}>
-          <Link to="#">Recipe {i}</Link>
+          <td>
+            <Link to="#">Recipe {i}</Link>
+          </td>
         </tr>
       );
     }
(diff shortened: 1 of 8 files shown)


src/pages/user/Recipes.js
```diff
@@ -9,8 +9,10 @@ export default class UserRecipes extends Component {
 
     for (let i = 0; i < 20; i++) {
       listItems.push(
-        <tr>
-          <Link to="#">Recipe {i}</Link>
+        <tr key={i}>
```
ayan4m1: I rolled this fix into #13, so GH should figure it out but you might need to rebase against master if it doesn't want to auto-merge for you.

ayan4m1:
Looks good - would you mind doing a rebase against master to resolve the conflicts? That should get rid of the test failures and then I think we can :shipit:

JosefBud:
> Looks good - would you mind doing a rebase against master to resolve the conflicts? That should get rid of the test failures and then I think we can :shipit:

I think it's good to go now!

ayan4m1:
Merge aborted in favor of #14

PR #16, opened 2019-07-16
Clean up Sass variable usage

ayan4m1:
* `$white` is already declared in Bootstrap, removed it
* [Apply](https://github.com/gusta-project/frontend/compare/master...ayan4m1:fix/sass-vars?expand=1#diff-4f4551829dd78334f41b7e816d80b553R97) two [changes](https://github.com/gusta-project/frontend/compare/master...ayan4m1:fix/sass-vars?expand=1#diff-4f4551829dd78334f41b7e816d80b553R106) to increase specificity of the selectors so that `!important` can be removed
* [Removing](https://github.com/gusta-project/frontend/compare/master...ayan4m1:fix/sass-vars?expand=1#diff-4f4551829dd78334f41b7e816d80b553R114) `!important` where it did not seem to be needed (tested in Chrome)
* [Switch](https://github.com/gusta-project/frontend/compare/master...ayan4m1:fix/sass-vars?expand=1#diff-4f4551829dd78334f41b7e816d80b553R128) to using the color variables instead of `#000` and `#fff`
* Similarly use [the gray mixin](https://github.com/gusta-project/frontend/compare/master...ayan4m1:fix/sass-vars?expand=1#diff-4f4551829dd78334f41b7e816d80b553R143) to simplify our gray palette

### DIFF
--- a/src/style.scss
+++ b/src/style.scss
@@ -94,7 +94,7 @@ hr {
 
 // Styling for dropdown links in the nav
 .dropdown-toggle {
-  color: #000 !important;
+  color: $black;
 }
 
 // Styling for the dropdown menu
@@ -126,8 +126,8 @@ a.active {
 .form-check-input {
   //-webkit-appearance: none;
   appearance: none;
-  background-color: #fff;
-  border: 1px solid #000;
+  background-color: $white;
+  border: 1px solid $black;
   width: 20px;
   height: 20px;
   border-radius: 0.3rem;
@@ -141,13 +141,14 @@ a.active {
 // Changing disabled buttons to grey
 .btn-primary.disabled,
 .btn-primary:disabled {
-  background-color: #777;
-  border-color: #333;
+  color: gray("800");
+  background-color: gray("300");
+  border-color: gray("800");
 }
 
 // Changing the border of any form fields
 .form-control {
-  border: 1px solid #000;// Style for "striped" tables
+  border: 1px solid $black; // Style for "striped" tables
   .table-striped tbody tr:nth-of-type(odd) {
     background-color: $lightest-blue;
   }
@@ -163,7 +164,7 @@ a.active {
 
 // Changing the background for the text next to an input field
 .input-group-text {
-  background-color: #eee;
+  background-color: gray("200");
 }
 
 // Changing the column width of the ingredient percent input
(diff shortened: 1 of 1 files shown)


src/style.scss
```diff
@@ -94,29 +94,28 @@ hr {
 }
 
 // Styling for dropdown links in the nav
-.dropdown-toggle {
-  color: #000 !important;
+.navbar-nav .nav-link.dropdown-toggle {
+  color: $black;
 }
```
JosefBud: On hover and focus, the dropdown link changes to Bootstrap's selector and its grey color. This will fix it:
```
  &:hover, &:focus {
    color: $black;
  }
```

src/style.scss
```diff
@@ -94,29 +94,28 @@ hr {
...
+    background-color: inherit;
+  }
 }
 
 // Styling for active navigation link
 a.active {
-  font-weight: 900 !important;
-  color: $teal !important;
+  font-weight: 900;
+  color: $teal;
```
JosefBud: The color is still overridden when the important tag is taken away. It can be missed if you don't click on the page to get rid of the focus styling from `.borderLeftRight`. I'd suggest the following change: 

`.navbar .navbar-nav .nav-link.active {` to replace `a.active {`

in conjunction with my suggestion on line 218 regarding the `.borderLeftRight` selectors

src/style.scss
```diff
@@ -219,7 +218,7 @@ h1.recipeTitle {
 .borderLeftRight {
```
JosefBud: To coincide with changes made to `a.active` on line 113, **all** instances of `.borderLeftRight` should be changed to `.navbar-nav .borderLeftRight.nav-link`

ayan4m1: OK, I refactored all of the selectors that had .borderLeftRight to be under one parent selector and then changed that parent selector in the way you described.

src/style.scss
ayan4m1: I found that it works as `.navbar-nav .nav-link.active` and we have a stylelint rule about max class chain being 2 - we can change the rule but for the time being that's what I updated it to.

src/style.scss
ayan4m1: Done!

src/style.scss
ayan4m1: Rule updated and selector has the class added, thanks!

JosefBud:
Let's get rid of the whole `.navbar-nav .nav-link.active {` selector on line 118 and just add `.active` to the `.borderLeftRight` nest starting at line 225. Get rid of the `!important` tag on 228 and 232 and we're good to go!

PR #24, opened 2019-07-21
Design/button class change

JosefBud:
My first instinct when adding the button animation was that all buttons should have it. While building the recipe page, I've realized that some buttons, like the favorite button or the rating stars, should not have the animation. So I've moved the animation away from the universal `.btn` class to its own `.button-animation` class and updated all of the appropriate buttons to include this new class.

### DIFF
--- a/src/pages/Calculator.js
+++ b/src/pages/Calculator.js
@@ -291,7 +291,7 @@ export default class Calculator extends Component {
           <h2>
             Flavor Stash
             <span>&nbsp;&nbsp;&nbsp;</span>
-            {this.state.ShowStash && (
+            {this.state.showStash ? (
               <Button
                 variant="info"
                 className="button-animation"
@@ -300,8 +300,7 @@ export default class Calculator extends Component {
               >
                 <span>Hide</span>
               </Button>
-            )}
-            {!this.state.ShowStash && (
+            ) : (
               <Button
                 variant="info"
                 className="button-animation"
--- a/src/pages/Calculator.js
+++ b/src/pages/Calculator.js
@@ -292,12 +292,22 @@ export default class Calculator extends Component {
             Flavor Stash
             <span>&nbsp;&nbsp;&nbsp;</span>
             {this.state.ShowStash && (
-              <Button variant="info" size="sm" onClick={e => this.hideStash(e)}>
+              <Button
+                variant="info"
+                className="button-animation"
+                size="sm"
+                onClick={e => this.hideStash(e)}
+              >
                 <span>Hide</span>
               </Button>
             )}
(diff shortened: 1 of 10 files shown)


src/pages/Calculator.js
```diff
@@ -337,12 +347,29 @@ export default class Calculator extends Component {
                               <td>{flavor.vendor.name}</td>
                               <td>{flavor.name}</td>
                               <td>
-                                <Button
-                                  onClick={this.addIngredient.bind(this, index)}
-                                  disabled={this.state.inuse[index]}
-                                >
-                                  <span>Add to Recipe</span>
-                                </Button>
+                                {this.state.inuse[index] && (
```
ayan4m1: You can use a ternary expression here for DRY purposes.

JosefBud: Good idea! Fixed in https://github.com/gusta-project/frontend/pull/24/commits/75fe7f03909ec7f2601a83956091ba63bdb87608

src/pages/Calculator.js
```diff
@@ -292,12 +292,22 @@ export default class Calculator extends Component {
             Flavor Stash
             <span>&nbsp;&nbsp;&nbsp;</span>
             {this.state.ShowStash && (
-              <Button variant="info" size="sm" onClick={e => this.hideStash(e)}>
+              <Button
```
ayan4m1: Ah, sorry I didn't notice, same ternary thing here.

src/pages/Login.js
```diff
@@ -81,8 +81,13 @@ export class Login extends Component {
...
               </Field>
-              <Button variant="primary" type="submit" disabled={submitting}>
-                Login
+              <Button
+                className="button-animation"
+                variant="primary"
+                type="submit"
+                disabled={submitting}
+              >
+                <span>Login</span>
```
ayan4m1: I just noticed that all of these buttons have `<span>` wrapping the text - if we add that wrapper Button component that can be one thing that goes away.

src/pages/user/Settings.js
```diff
@@ -98,7 +98,10 @@ export default class UserSettings extends Component {
                   </Form.Label>
                 </div>
                 <InputGroup.Prepend>
-                  <Button style={{ margin: '0px 10px' }}>
+                  <Button
+                    className="button-animation"
+                    style={{ margin: '0px 10px' }}
```
ayan4m1: Rargh blargh inline styles.

src/pages/Login.js
JosefBud: That would be awesome, I definitely think it's annoying to have to `<span>` the text inside every button

PR #31, opened 2019-08-11
Refactoring Sass variables & stylelint warning fix

JosefBud:
https://github.com/gusta-project/frontend/commit/d765c2be0d39d0c885c5eb7534ff1f7e92b0a0a1 Moved the color variables to a new `_variables.scss` file, and imported back into `style.scss`.

https://github.com/gusta-project/frontend/commit/4e7c3f41b85128bd4f26fe3dad98b49d770b1a76 Added breakpoint sizes to `_variables.scss` and changed all breakpoints in `style.scss` to use these new variables.

https://github.com/gusta-project/frontend/commit/297908259212446e738bfc2d7dc9b2ab571f1bf2 Fixed the selectors for striped tables to make stylelint happy. There are now no longer any warnings from stylelint 🎉 .

### DIFF
--- a/src/style.scss
+++ b/src/style.scss
@@ -49,13 +49,13 @@ hr {
 // it changes the width of the container and the width of the other elements around it
 // Switching the container to "fluid" didn't fix it, so this was unfortunately the best way I could
 // find to solve the issue; by fixing the container width for medium screens and up
-@media (min-width: $min-md) and (max-width: $max-lg) {
+@media (min-width: $min-md) and (max-width: $max-md) {
   .container {
     width: 800px;
   }
 }
 
-@media (min-width: $min-lg) and (max-width: $max-xl) {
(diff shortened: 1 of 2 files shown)


src/style.scss
```diff
@@ -57,19 +49,19 @@ hr {
 // it changes the width of the container and the width of the other elements around it
 // Switching the container to "fluid" didn't fix it, so this was unfortunately the best way I could
 // find to solve the issue; by fixing the container width for medium screens and up
-@media (min-width: 768px) and (max-width: 991px) {
+@media (min-width: $md) and (max-width: $lg) {
```
ayan4m1: Don't you run into conflicting breakpoints unless you do `calc($lg - 1)` ?

JosefBud: @ayan4m1 just tested it, and it doesn't seem to make a difference but I can change that if you think it would be best practice to do so! 😄

PR #32, opened 2019-08-14
Recipe viewer page

JosefBud:
To view the recipe page, navigate to `localhost:9000/recipe?id=[1-6]` and pick any number in that range... The recipe with the ID `6` uses a couple of flavors that are also in the mock data flavor stash, to see the "is it in your stash?" check at work.

**Changes:**
- Several style changes
- Added favorite, rating, and "make this recipe" buttons
- Added CSS classes specifically for favorite and rating buttons
- Added a function to check recipe flavors against stash flavors, with a fake recipe added (ID 6) to test it
- Added mock recipe tags
- Added mock recipe version
- Added toasts for "favoriting" a recipe
- Updated the page test

There was one change that affects more than the recipe page, and that is https://github.com/gusta-project/frontend/commit/b3a290d2afd64c8f2fe017691fedb0d807fb8db6 where I took the button border and background color and moved it from the `.button-animation` class to the global `.btn` class.


ayan4m1:
Getting some errors in the tests (on CircleCI as well)

```
    console.error node_modules/prop-types/checkPropTypes.js:20
      Warning: Failed prop type: The prop `actions` is marked as required in `Recipe`, but its value is `undefined`.
          in Recipe
    console.error node_modules/@fortawesome/react-fontawesome/index.js:325
      Could not find icon { prefix: 'far', iconName: 'heart' }
    console.error node_modules/@fortawesome/react-fontawesome/index.js:325
      Could not find icon { prefix: 'far', iconName: 'star' }
    console.error node_modules/@fortawesome/react-fontawesome/index.js:325
      Could not find icon { prefix: 'far', iconName: 'star' }
    console.error node_modules/@fortawesome/react-fontawesome/index.js:325
```

I'll follow up with comments in the PR about what I think is going on there.

src/pages/Recipe.js
```diff
@@ -2,25 +2,84 @@ import PropTypes from 'prop-types';
...
+  Row,
+  Col,
+  Table,
+  Button,
+  ButtonGroup,
+  Card
+} from 'react-bootstrap';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+
+import flavorStash from '../data/flavorstash.json';
```
ayan4m1: We can fetch this data from the API now, but that can be done in a separate PR.

src/pages/Recipe.js
```diff
@@ -2,25 +2,84 @@ import PropTypes from 'prop-types';
...
+export class Recipe extends Component {
   static propTypes = {
+    actions: PropTypes.object.isRequired,
     location: PropTypes.object
   };
 
-  constructor(...args) {
-    super(...args);
+  constructor(props) {
+    super(props);
```
ayan4m1: 👍

src/pages/Recipe.js
```diff
@@ -2,25 +2,84 @@ import PropTypes from 'prop-types';
...
+  }
+
+  handleFavoriteClick() {
+    if (this.state.favorited) {
+      this.setState({
+        favorited: false,
+        favoriteIcon: ['far', 'heart']
+      });
+
+      this.props.actions.popToast({
```
ayan4m1: Destructure actions off this.props at the top of this method.

src/pages/Recipe.js
```diff
@@ -35,37 +94,189 @@ export default class Recipe extends Component {
       recipe.flavorTotal += flavor.percent;
     });
 
+    if (recipe.notes) {
+      const newNotes = recipe.notes
+        .toString()
+        .split('\n')
+        .map((line, i) => {
+          return <p key={i}>{line}</p>;
```
ayan4m1: Using `key={i}` might not be viable here, since if the recipe changes the key will not and React may fail to update the notes text in strange ways. Build a string like `` key={`${id}${i}`} `` and you should be good.

src/pages/Recipe.js
```diff
@@ -35,37 +94,189 @@ export default class Recipe extends Component {
...
+    });
+
+    recipe.tags = newTags;
+
+    this.setState(recipe);
+  }
+
+  compareWithStash() {
+    // eslint-disable-next-line
+    console.log(this.state);
```
ayan4m1: Yank this

src/pages/Recipe.js
ayan4m1: It _might_ work in this scenario, but the general guidance (this is from React docs)

https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318

src/pages/Recipe.js
```diff
@@ -35,37 +94,189 @@ export default class Recipe extends Component {
...
+
     this.setState(recipe);
   }
 
+  renderRatingButtons(ratingNum) {
+    const buttons = [];
+
+    let ratingKey = 1;
+
+    for (let i = 0; i < ratingNum; i++) {
```
ayan4m1: Just stride over `0 ... 5` once and push one type of button or the other - the markup is very similar so this could be more DRY.

src/pages/Recipe.js
```diff
@@ -35,37 +94,189 @@ export default class Recipe extends Component {
...
+        >
+          <span>
+            <FontAwesomeIcon icon={['far', 'star']} />
+          </span>
+        </Button>
+      );
+      ratingKey++;
+    }
+
+    if (ratingKey === 6) {
```
ayan4m1: Does this return undefined if ratingKey is not 6 at the end?

src/pages/Recipe.js
```diff
@@ -35,37 +94,189 @@ export default class Recipe extends Component {
...
+              </span>
+            </Button>
+          </Col>
+          <Col xs="auto">{this.renderRatingButtons(this.state.rating)}</Col>
+        </Row>
+        <hr />
+        <Row>
+          <Col lg={{ span: 8, offset: 2 }} xs={{ span: 12 }}>
+            <Card className="text-center">
+              <Card.Header>{this.state.name}</Card.Header>
```
ayan4m1: Destructure all these state properties at the top of this method.

src/pages/Recipe.js
```diff
@@ -77,36 +288,42 @@ export default class Recipe extends Component {
...
-            </Col>
-          </Row>
-        )}
       </Container>
     );
   }
 }
+
+const mapDispatchToProps = dispatch => ({
+  actions: bindActionCreators(
```
ayan4m1: This also passes prettier rules and is more concise:

```js
const mapDispatchToProps = dispatch => ({
  actions: bindActionCreators({ ...actions }, dispatch)
});
```

src/pages/Recipe.js
JosefBud: Yikes.

src/pages/Recipe.test.js
```diff
@@ -1,9 +1,15 @@
 import React from 'react';
+import { Provider } from 'react-redux';
 import renderer from 'react-test-renderer';
+import configureStore from 'redux-mock-store';
 
-import Recipe from './Recipe';
+import { initialState } from 'reducers/application';
```
ayan4m1: Use `import { initialState } from 'reducers';` and pass that in to the mockStore instead, this gets you the whole state rather than having to assemble it as we add reducers.

src/pages/Recipe.js
```diff
@@ -35,37 +94,189 @@ export default class Recipe extends Component {
...
+        .map((line, i) => {
+          return <p key={i}>{line}</p>;
+        });
+
+      recipe.notes = newNotes;
+    }
+
+    const newTags = recipe.tags.map((tag, i) => {
+      return (
+        <a className="link--tags" key={i} href="/">
```
ayan4m1: Try [this out](https://getbootstrap.com/docs/4.3/components/badge/#links)

src/pages/Recipe.js
```diff
@@ -35,37 +94,189 @@ export default class Recipe extends Component {
...
+              <caption className="text-center">
+                {this.state.maxvg ? (
+                  <span>Max VG |&nbsp;</span>
+                ) : (
+                  <span>
+                    {this.state.pg}% PG / {this.state.vg}% VG |&nbsp;
+                  </span>
+                )}
+                {this.state.shakeNVape ? (
+                  <span>Shake & Vape |&nbsp;</span>
```
ayan4m1: `&amp;`

src/pages/Recipe.test.js
```diff
@@ -15,7 +21,11 @@ describe('<Recipe />', () => {
   });
 
   it('renders correctly', () => {
-    const component = renderer.create(<Recipe {...props('?id=1')} />);
+    const component = renderer.create(
+      <Provider store={store}>
+        <Recipe {...props('?id=1')} />
```
ayan4m1: You need to pass actions to this guy, sorry if I was unclear earlier when I said not to pass props to it - the only props you'd want to pass would be special props like `location` or `actions` that get injected by a HOC.

src/pages/Recipe.js
JosefBud: I think I did this because I was running into issues with all 5 rating buttons rendering, so I added this to basically loop through and say once the loop is done, render the button group. Changed this as well though.

src/pages/Recipe.js
```diff
@@ -35,78 +97,226 @@ export default class Recipe extends Component {
...
+        .map((line, i) => {
+          return <Card.Text key={`${recipe.id}${i}`}>{line}</Card.Text>;
+        });
+
+      recipe.notes = newNotes;
+    }
+
+    const newTags = recipe.tags.map((tag, i) => {
+      return (
+        <a key={i} href="/">
```
ayan4m1: This guy needs the same fix as above - in general any time you're using `.map()` you need to pass it a key that changes if and only if the actual data being stored changes.

src/pages/Recipe.js
```diff
@@ -35,78 +97,226 @@ export default class Recipe extends Component {
...
+  }
+
+  renderRatingButtons(ratingNum) {
+    const buttons = [];
+
+    for (let i = 0; i < 5; i++) {
+      buttons.push(
+        <Button
+          className="rating-button button--recipe"
+          key={i}
```
ayan4m1: This is likely an instance where using `key={i}` is OK.

src/pages/Recipe.js
```diff
@@ -35,78 +97,226 @@ export default class Recipe extends Component {
...
+          </span>
+        </Button>
+      );
+    }
+
+    return <ButtonGroup>{buttons}</ButtonGroup>;
+  }
+
   render() {
+    const {
```
ayan4m1: 💥

PR #35, opened 2019-08-19
Update React Bootstrap package

JosefBud:
Updating from beta 9 to beta 11 to fix issues with overlay triggers & popovers


JosefBud:
> Good fix @daviddyess and @JosefBud !

@daviddyess gets all the credit for this one, I'm only a mere vessel

PR #43, opened 2019-09-02
Administrative Dashboard

daviddyess:
Administrative dashboard for managing site content and settings:

- Site statistics
- View, Add, Remove, Edit, Assign: Roles
- View Users
- View Ingredients (only flavors for now)
- View Database migration status


src/components/Dashboard/Flavors.js
```diff
@@ -0,0 +1,81 @@
...
+import { actions as flavorsActions } from 'reducers/flavors';
+import { getAllFlavors } from 'selectors/flavors';
+
+export class Flavors extends Component {
+  static propTypes = {
+    actions: PropTypes.object.isRequired,
+    opt: PropTypes.object.isRequired,
+    flavors: PropTypes.array
+  };
+  constructor(props) {
```
ayan4m1: No need for the constructor if all you're doing is calling `super`

src/components/Dashboard/Flavors.js
```diff
@@ -0,0 +1,81 @@
...
+  }
+}
+
+const mapStateToProps = state => ({
+  flavors: getAllFlavors(state)
+});
+
+const mapDispatchToProps = dispatch => ({
+  actions: bindActionCreators(
+    {
```
ayan4m1: This is something I started doing that I had fixed in the recipe-editor branch - no need to spread it if there's only one set of actions you need to map.

src/components/Dashboard/Flavors.test.js
```diff
@@ -0,0 +1,30 @@
...
+    header: true,
+    title: false,
+    style: {}
+  };
+  const mockStore = configureStore();
+  const store = mockStore({ flavors: initialState });
+  const RoutedFlavors = withMemoryRouter(ConnectedFlavors);
+
+  it('renders correctly', () => {
+    const component = renderer.create(
```
ayan4m1: I don't consider this gating, but if you don't need the reference to `component` then you can just do `const tree = renderer.create(<Blah />).toJSON();` and then the expect is a little cleaner.

src/components/Dashboard/Home.js
```diff
@@ -0,0 +1,103 @@
...
+          </Card>
+          <Card bg="warning" text="white">
+            <Card.Header>Recipes: {recipes}</Card.Header>
+            <Card.Body>
+              <h2>Tags: {recipeTags}</h2>
+            </Card.Body>
+          </Card>
+        </CardColumns>
+        <a href="#home" onClick={this.refresh} title="Refresh Stats">
+          <FontAwesomeIcon icon={faSync} />
```
ayan4m1: Refactor this so that `faSync` is added to the library in icons.js and then use `"sync"` here.

src/components/Dashboard/Layout.js
```diff
@@ -0,0 +1,36 @@
+import PropTypes from 'prop-types';
+import { Helmet } from 'react-helmet';
+import React, { Component } from 'react';
+import { Card } from 'react-bootstrap';
+export class Layout extends Component {
```
ayan4m1: I am wondering if we should refactor these to `src/layouts` - so that we don't have a bunch of classes all called Layout. What do you think?

src/components/Dashboard/Layout.js
```diff
@@ -0,0 +1,36 @@
...
+  static propTypes = {
+    children: PropTypes.any.isRequired,
+    header: PropTypes.string,
+    options: PropTypes.object,
+    pageTitle: PropTypes.string,
+    subTitle: PropTypes.any
+  };
+  constructor(props) {
```
ayan4m1: Unnecessary constructor.

src/pages/Dashboard.js
```diff
@@ -0,0 +1,65 @@
...
+
+import { actions as dashboardActions } from 'reducers/dashboard';
+import { getDashboardComponent } from 'selectors/dashboard';
+
+export class Dashboard extends Component {
+  static propTypes = {
+    actions: PropTypes.object.isRequired
+  };
+
+  constructor(props) {
```
ayan4m1: Unnecessary constructor.

src/components/Dashboard/Users.js
```diff
@@ -0,0 +1,86 @@
...
+import { actions as usersActions } from 'reducers/users';
+import { getAllUsers } from 'selectors/users';
+
+export class Users extends Component {
+  static propTypes = {
+    actions: PropTypes.object.isRequired,
+    users: PropTypes.array,
+    opt: PropTypes.object
+  };
+  constructor(props) {
```
ayan4m1: Unnecessary constructor.

src/components/Dashboard/Roles.js
```diff
@@ -0,0 +1,123 @@
...
+import { actions as rolesActions } from 'reducers/roles';
+import { getAllRoles } from 'selectors/roles';
+
+export class Roles extends Component {
+  static propTypes = {
+    actions: PropTypes.object.isRequired,
+    opt: PropTypes.object,
+    roles: PropTypes.array
+  };
+  constructor(props) {
```
ayan4m1: Unnecessary constructor.

src/components/Dashboard/Migrations.js
```diff
@@ -0,0 +1,82 @@
...
+import { actions as dashboardActions } from 'reducers/dashboard';
+import { getMigrations } from 'selectors/dashboard';
+
+export class Migrations extends Component {
+  static propTypes = {
+    actions: PropTypes.object.isRequired,
+    opt: PropTypes.object.isRequired,
+    migrations: PropTypes.array
+  };
+  constructor(props) {
```
ayan4m1: Unnecessary constructor.

src/components/Dashboard/Menu.js
```diff
@@ -0,0 +1,117 @@
...
+  faPlus,
+  faUsers,
+  faUsersCog,
+  faUserShield,
+  faTachometerAlt
+} from '@fortawesome/free-solid-svg-icons';
+// import {  } from '@fortawesome/free-regular-svg-icons';
+
+export class Menu extends Component {
+  constructor(props) {
```
ayan4m1: Unnecessary constructor.

src/components/Dashboard/Main.js
```diff
@@ -0,0 +1,111 @@
...
+
+import { actions as dashboardActions } from 'reducers/dashboard';
+import { getDashboardComponent } from 'selectors/dashboard';
+
+class Main extends Component {
+  static propTypes = {
+    actions: PropTypes.object.isRequired,
+    dashboardComponent: PropTypes.object.isRequired
+  };
+  constructor(props) {
```
ayan4m1: Unnecessary constructor.

src/components/Dashboard/Link.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  render() {
+    const { to, name, item, className } = this.props;
+
+    return (
+      <Fragment>
+        <a
+          href={to}
+          onClick={e => this.select(name, item, e)}
```
ayan4m1: `e` is passed as a third argument that is not present in the `select` method signature. I think `() =>this.select(name, item)` works.

src/components/Dashboard/Link.js
```diff
@@ -0,0 +1,68 @@
...
+          className={className}
+        >
+          {this.props.children}
+        </a>
+      </Fragment>
+    );
+  }
+}
+
+// export default Dashboard;
```
ayan4m1: Stray code.

src/pages/Dashboard.js
```diff
@@ -0,0 +1,65 @@
...
+          <Col xs={12} md={9}>
+            <Main />
+          </Col>
+        </Row>
+      </Container>
+    );
+  }
+}
+
+// export default Dashboard;
```
ayan4m1: Don't think this is needed anymore.

src/components/Dashboard/index.js
```diff
@@ -0,0 +1,39 @@
+import LinkExport from './Link';
```
ayan4m1: Take a look [here](https://github.com/react-bootstrap/react-bootstrap/blob/master/src/index.js) for an example of a more concise way to do this.

src/pages/Dashboard.test.js
```diff
@@ -0,0 +1,24 @@
...
+import renderer from 'react-test-renderer';
+import configureStore from 'redux-mock-store';
+
+import { initialState } from 'reducers/dashboard';
+import ConnectedDashboard from './Dashboard';
+import { withMemoryRouter } from 'utils';
+
+describe('<Dashboard />', () => {
+  const mockStore = configureStore();
+  const store = mockStore({ dashboard: initialState });
```
ayan4m1: To fill the snapshot with actual data, pass a modified initialState into mockStore. Otherwise the snapshot is just a test that there is no rendering error, not a test of whether or not the component can display correctly with data.

src/components/Dashboard/Flavors.js
daviddyess: Removed

src/components/Dashboard/Layout.js
daviddyess: Sounds like a good idea, I'll move it to there

src/components/Dashboard/Layout.js
daviddyess: Removed

src/pages/Dashboard.js
daviddyess: Removed

src/components/Dashboard/Users.js
daviddyess: Removed

src/components/Dashboard/Roles.js
daviddyess: removed

src/components/Dashboard/Migrations.js
daviddyess: removed

src/components/Dashboard/Menu.js
daviddyess: removed

src/components/Dashboard/Main.js
daviddyess: removed

src/components/Dashboard/Layout.js
daviddyess: Moved to layouts

src/components/Dashboard/index.js
daviddyess: Updated

src/components/Dashboard/Flavors.test.js
daviddyess: I missed this earlier, probably quite a few more now. I'll correct them (so you don't have to repeat this comment 30 times ;)

src/components/Dashboard/Link.js
daviddyess: fixed

src/components/Dashboard/Link.js
daviddyess: fixed

src/pages/Dashboard.js
daviddyess: fixed

src/pages/Dashboard.test.js
daviddyess: initialState loads the dashboard homepage, I am currently testing data in Main.js

src/components/Dashboard/Home.js
daviddyess: All icons have been added to library (hopefully)

src/components/Dashboard/Flavors.js
daviddyess: This should be fixed everywhere now

src/components/Dashboard/Flavors.js
```diff
@@ -0,0 +1,73 @@
...
+import { Table } from 'react-bootstrap';
+import { DashboardLayout as Layout } from 'components/Dashboard/';
+
+import { actions as flavorsActions } from 'reducers/flavors';
+import { getAllFlavors } from 'selectors/flavors';
+
+export class Flavors extends Component {
+  static propTypes = {
+    actions: PropTypes.object.isRequired,
+    opt: PropTypes.object.isRequired,
```
ayan4m1: Would you mind calling this `options` or `layoutOptions` maybe?

src/components/Dashboard/Main.js
```diff
@@ -0,0 +1,103 @@
...
+
+import { actions as dashboardActions } from 'reducers/dashboard';
+import { getDashboardComponent } from 'selectors/dashboard';
+
+class DashboardMain extends Component {
+  static propTypes = {
+    actions: PropTypes.object.isRequired,
+    dashboardComponent: PropTypes.object.isRequired
+  };
+  display() {
```
ayan4m1: You can make this a getter, but you might as well just fold it into the render method since render isn't doing anything else.

src/components/Dashboard/Main.js
```diff
@@ -0,0 +1,103 @@
...
+    const { dashboardComponent } = this.props;
+    const { name, item } = dashboardComponent;
+    const defaultOpts = {
+      border: false,
+      header: true,
+      title: false,
+      style: {}
+    };
+
+    switch (name) {
```
ayan4m1: You could export a const (effectively an enum) with the component names and then create a map like

```jsx
export const ComponentNames = {
  HOME: 'Home'
};
export const Components = {
  ComponentNames.HOME: <Home />,
};
```

Then to render

```jsx
render() {
  const { dashboardComponent } = this.props;
  const { name, item } = dashboardComponent;
  const Component = Components[name];

  return <Component {...this.props} />;
}
```

I don't mind the switch but this might be a case where having the enum is useful. No need to change it unless you want to.

src/layouts/DashboardLayout.js
```diff
@@ -0,0 +1,33 @@
+import PropTypes from 'prop-types';
+import { Helmet } from 'react-helmet';
+import React, { Component } from 'react';
+import { Card } from 'react-bootstrap';
+export class DashboardLayout extends Component {
```
ayan4m1: Mind adding a newline between the imports and export?

src/pages/Dashboard.js
```diff
@@ -0,0 +1,62 @@
...
+  }
+}
+
+const mapStateToProps = state => ({
+  dashboardComponent: getDashboardComponent(state)
+});
+
+const mapDispatchToProps = dispatch => ({
+  actions: bindActionCreators(
+    {
```
ayan4m1: No need to spread a single argument, just pass it directly

src/reducers/dashboard.test.js
```diff
@@ -0,0 +1,135 @@
...
+  const dashboardComponent = {
+    name: 'Home',
+    item: null
+  };
+  const name = 'Home';
+  const item = null;
+  const migrations = [{ db: true }];
+  const stats = [{ users: true }];
+  const error = { message: 'Failed' };
+  /* const toast = {
```
ayan4m1: Nix this if it's not needed.

src/reducers/flavors.js
```diff
@@ -0,0 +1,48 @@
+import { buildActions } from 'utils';
+
+export const types = buildActions('flavors', [
```
ayan4m1: My recipe editor PR also adds one of these, but I'll rebase mine off yours and deal with the conflicts.

src/reducers/index.js
```diff
@@ -5,10 +5,29 @@ import {
   initialState as applicationState
 } from './application';
 
+import {
+  reducer as dashboard,
+  initialState as dashboardState
+} from './dashboard';
+
+import { reducer as flavors, initialState as flavorsState } from './flavors';
```
ayan4m1: No need for newlines between these imports

src/components/Dashboard/Migrations.js
```diff
@@ -0,0 +1,74 @@
...
+import { Table } from 'react-bootstrap';
+import { DashboardLayout as Layout } from 'components/Dashboard/';
+
+import { actions as dashboardActions } from 'reducers/dashboard';
+import { getMigrations } from 'selectors/dashboard';
+
+export class Migrations extends Component {
+  static propTypes = {
+    actions: PropTypes.object.isRequired,
+    opt: PropTypes.object.isRequired,
```
ayan4m1: Mind changin this to `options` or `layoutOptions`?

src/components/Dashboard/Role/Users.js
```diff
@@ -0,0 +1,105 @@
...
+      <Layout
+        pageTitle="Role Users - Dashboard"
+        header={`Roles > ${name}`}
+        options={opt}
+      >
+        <FontAwesomeIcon icon="chevron-left" /> &nbsp;
+        <DashLink to="#roles" name="Roles">
+          Back
+        </DashLink>
+        {!roleUsers && <Alert>No Users Assigned to this Role!</Alert>}
```
ayan4m1: You should use a ternary here or else odd behavior can occur where both are displayed for a brief moment.

src/components/Dashboard/Role/DeleteUser.test.js
```diff
@@ -0,0 +1,75 @@
...
+        )
+        .toJSON()
+    ).toMatchSnapshot();
+  });
+
+  it('can handleSubmit', () => {
+    const component = renderer.create(
+      <Provider store={store}>
+        <RoleDeleteUser
+          actions={actions}
```
ayan4m1: You could declare

```jsx
const props = {
  actions,
  opt: defaultOPts,
  userId,
  roleId,
  name
};
```

and then `<RoleDeleteUser {...props} />`

src/components/Dashboard/Role/Delete.test.js
```diff
@@ -0,0 +1,68 @@
...
+          </Provider>
+        )
+        .toJSON()
+    ).toMatchSnapshot();
+  });
+
+  it('can handleSubmit', () => {
+    const component = renderer.create(
+      <Provider store={store}>
+        <RoleDelete
```
ayan4m1: You could declare these props up at the top of the describe block and then spread them here.

src/components/Dashboard/Flavors.js
daviddyess: Should be fixed in all locations, also changed defaultOpts to defaultLayoutOptions where applicable. opt is now layoutOptions

src/components/Dashboard/Main.js
daviddyess: Fixed

src/layouts/DashboardLayout.js
daviddyess: fixed

src/pages/Dashboard.js
daviddyess: fixed

src/components/Dashboard/Role/AddUser.js
```diff
@@ -0,0 +1,141 @@
...
+    this.state = { userId: '0', submitting: false };
+
+    this.handleChange = this.handleChange.bind(this);
+    this.handleSubmit = this.handleSubmit.bind(this);
+  }
+
+  componentDidMount() {
+    const { actions } = this.props;
+
+    actions.requestUsers();
```
ayan4m1: This is an instance where we'd ideally have a boolean `loaded` prop in the redux state tree that is set to true on `_SUCCESS` - to cut down on network calls if the user is moving between pages.

src/components/Dashboard/Role/AddUser.js
```diff
@@ -0,0 +1,141 @@
...
+              onChange={e => this.handleChange(e)}
+              value={userId}
+            >
+              <option value="0" key="x">
+                Choose a User
+              </option>
+              {users.map((user, index) => {
+                return (
+                  <option value={user.id} key={index}>
+                    {user.emailAddress}
```
ayan4m1: Might be nice to show the UserProfile username here in parens so we could convert it to a typeahead at some point in the future.

src/components/Dashboard/Role/AddUser.js
```diff
@@ -0,0 +1,141 @@
...
+                  </option>
+                );
+              })}
+            </Form.Control>
+          </Form.Group>
+          {userId !== '0' && (
```
ayan4m1: Usually when I need to make one assertion in multiple JSX expressions I throw it up at the top of the render method as a const, makes things easier to refactor:

```jsx
const isUserIdSelected = userId !== '0';
```

And if the check is more complicated than just one or two clauses I'll throw it into a getter like:

```jsx
get userIdSelected() {
  const { something, another, third } = this.props;
  return (something && another) || (third && !something);
}

...

{this.userIdSelected ? <Something /> : <TheOtherThing />}
```

src/components/Dashboard/Role/AddUser.test.js
```diff
@@ -0,0 +1,70 @@
...
+    title: false,
+    style: {}
+  };
+  const mockStore = configureStore();
+  const store = mockStore({
+    roles: initialState,
+    dashboard: dashboardInitialState,
+    users: usersInitialState
+  });
+  /* const actions = {
```
ayan4m1: Get rid of this unless necessary.

src/components/Dashboard/Users.js
```diff
@@ -0,0 +1,85 @@
...
+              return (
+                <tr key={index}>
+                  <td>{user.id}</td>
+                  <td>{user.emailAddress}</td>
+                  <td>{user.activationCode === null ? 'Yes' : 'No'}</td>
+                  <td>
+                    <Link to="/user/profile">Profile</Link>
+                    &nbsp; | &nbsp;
+                    <DashLink
+                      to={'#user/' + user.id + '/roles'}
```
ayan4m1: Could use string interpolation here.

src/components/Dashboard/Users.js
```diff
@@ -0,0 +1,85 @@
...
+              <th>Options</th>
+            </tr>
+          </thead>
+          <tbody>
+            {users.map((user, index) => {
+              return (
+                <tr key={index}>
+                  <td>{user.id}</td>
+                  <td>{user.emailAddress}</td>
+                  <td>{user.activationCode === null ? 'Yes' : 'No'}</td>
```
ayan4m1: Maybe use a check/X icon with Yes and No as alt text?

src/reducers/dashboard.js
```diff
@@ -0,0 +1,121 @@
...
+  requestStatsFailure
+};
+
+export const initialState = {
+  dashboardComponent: {
+    name: 'Home',
+    item: null
+  },
+  migrations: [],
+  stats: []
```
ayan4m1: Isn't stats an object normally? Should probably `{}` to prevent prop-types warnings.

src/sagas/flavors.js
```diff
@@ -0,0 +1,68 @@
...
+  // take,
+  // takeEvery,
+  // select
+} from 'redux-saga/effects';
+
+import request from 'utils/request';
+import { actions, types } from 'reducers/flavors';
+import { actions as appActions } from 'reducers/application';
+
+// snake_cased variables here come from RFC 6749
```
ayan4m1: This comment and the one below it are likely from a different file and unrelated.

src/sagas/flavors.js
```diff
@@ -0,0 +1,68 @@
+// import dayjs from 'dayjs';
+// import nanoid from 'nanoid';
+import {
+  all,
+  call,
+  put,
+  takeLatest
+  // delay,
```
ayan4m1: Remove these imports unless they're being used.

src/sagas/dashboard.test.js
```diff
@@ -0,0 +1,168 @@
+import { all, put, call } from 'redux-saga/effects';
+
+import request from 'utils/request';
+import { actions } from 'reducers/dashboard';
+import { actions as appActions } from 'reducers/application';
+import saga, { watchers, workers } from './dashboard';
+
+/* eslint-disable camelcase */
```
ayan4m1: Is this needed?

src/sagas/index.js
```diff
@@ -1,7 +1,15 @@
 import { fork } from 'redux-saga/effects';
 
 import application from './application';
+import dashboard from './dashboard';
+import flavors from './flavors';
+import roles from './roles';
+import users from './users';
 
 export default function* saga() {
   yield fork(application);
```
ayan4m1: Try this out:

```jsx
yield all([application, dashboard, flavors, roles, users].map(fork));
```

src/sagas/index.test.js
```diff
@@ -2,12 +2,25 @@ import { fork } from 'redux-saga/effects';
...
 
 describe('index saga', () => {
   it('forks all sagas', () => {
     const gen = saga();
-    const result = gen.next();
+
+    let result = gen.next();
 
     expect(result.value).toEqual(fork(application));
+    result = gen.next();
```
ayan4m1: This can become

```jsx
expect(result.value).toEqual(all([application, flavor, recipe].map(fork)));
```

src/selectors/flavors.js
```diff
@@ -0,0 +1,8 @@
+import { createSelector } from 'reselect';
+
+export const getFlavors = state => state.flavors;
+
+export const getAllFlavors = createSelector(
+  getFlavors,
+  flavors => flavors.flavors
```
ayan4m1: Usually when I see a word duplicated like this in one of my selectors I try to refactor the redux state so that it reads cleanly, like `flavors.list` - not saying you should change it to that or that we should use `.list` as a pattern, just wanted to mention it.

src/selectors/users.js
```diff
@@ -0,0 +1,19 @@
+// import dayjs from 'dayjs';
```
ayan4m1: Nix this import unless needed.

src/sagas/users.js
```diff
@@ -0,0 +1,148 @@
...
+      url: `/user/${userId}/roles`,
+      method: 'GET'
+    };
+    const result = yield call(request.execute, { endpoint });
+
+    // update user roles in state or throw an error
+    if (result.success) {
+      const {
+        response: { data }
+      } = result;
```
ayan4m1: Just for reference, any expensive transformations of these results should go here rather than in component methods whenever possible.

src/sagas/roles.js
```diff
@@ -0,0 +1,334 @@
...
+      appActions.popToast({
+        title: 'Error',
+        icon: 'times-circle',
+        message
+      })
+    );
+  }
+}
+
+function* createRoleWorker({ name }) {
```
ayan4m1: There's a pretty big code coverage gap in this file.

src/components/Dashboard/Users.js
```diff
@@ -47,14 +48,20 @@ export class Users extends Component {
...
               return (
                 <tr key={index}>
-                  <td>{user.id}</td>
+                  <td className="text-center">{user.id}</td>
                   <td>{user.emailAddress}</td>
-                  <td>{user.activationCode === null ? 'Yes' : 'No'}</td>
-                  <td>
+                  <td className="text-center">
+                    {user.activationCode === null ? (
+                      <FontAwesomeIcon icon="check" color="green" title="Yes" />
```
ayan4m1: It's up to you but one thing I sometimes do with things like this is to split the icons out into separate getter methods and then just `{user.activationCode === null ? this.yesIcon : this.noIcon}`

src/pages/Dashboard.test.js
ayan4m1: Gotcha, makes sense.

src/reducers/users.js
```diff
@@ -57,6 +63,7 @@ const requestUserRolesFailure = error => ({
 });
 
 export const actions = {
+  resetLoaded,
```
ayan4m1: How about a setLoaded instead that takes a new object to use in place of the existing `loaded` object?

That way we can avoid [this code](https://github.com/mixnjuice/frontend/pull/43/files/a8c6ae553547e9c9694aa1d3d49c9324c2218e37..99e5ff2b8d1e9869d58c62f119846312339a7ef5#diff-3f5cc67cba0731d032221ec5bcffd3bdR90) which is going to be hard to scale.

src/selectors/flavors.js
daviddyess: I refactored this to flavors.collection

src/sagas/roles.js
daviddyess: Added coverage for this file

src/reducers/users.js
daviddyess: I refactored to use a `users.cache`, which stores the query data by page/limit, then `users.collection` holds the currently displayed data. An API request is only made if the request isn't already in the `users.cache`. This scales without the need for the problem area you identified.

src/components/Dashboard/Role/AddUser.js
daviddyess: Updated

src/components/Dashboard/Users.js
daviddyess: I added this

src/sagas/dashboard.test.js
daviddyess: removed

src/reducers/dashboard.js
daviddyess: Updated

src/components/Dashboard/Role/AddUser.js
daviddyess: Added usernames

src/components/Dashboard/Flavors.js
```diff
@@ -2,34 +2,104 @@ import PropTypes from 'prop-types';
...
     const { actions } = this.props;
 
-    actions.requestFlavors();
+    actions.requestFlavors({ page: 1, limit: 20 });
+  }
+
+  pagerCounter() {
+    const { pages } = this.props.pager;
+
+    let i = 1;
```
ayan4m1: This could be

```jsx
return [...Array(pages).keys()].map(value => value + 1);
```

src/components/Dashboard/Flavors.js
```diff
@@ -2,34 +2,104 @@ import PropTypes from 'prop-types';
...
+    while (i <= pages) {
+      pager[i] = i;
+      i++;
+    }
+    return pager;
+  }
+
+  changePage(page) {
+    const { actions, pager } = this.props;
+
```
ayan4m1: Use destructuring here, like `const { target: { value } } = page;`

src/components/Dashboard/Flavors.test.js
```diff
@@ -29,4 +45,40 @@ describe('Dashboard <Flavors />', () => {
...
+      <Provider store={store}>
+        <RoutedFlavors actions={actions} layoutOptions={defaultLayoutOptions} />
+      </Provider>
+    );
+    const { instance } = component.root.findByType(Flavors);
+
+    expect(instance).toBeDefined();
+    instance.changePage(page);
+  });
```
ayan4m1: There should be some expectations made after to assert that [setState and actions.requestFlavors](https://github.com/mixnjuice/frontend/pull/43/files/99e5ff2b8d1e9869d58c62f119846312339a7ef5..6672cb715193e110eb18a65755d636f04a554ab4#diff-e9aa3e7b4635a24e56bf93f151e5f8d6R50) are being called.

src/components/Dashboard/Users.js
```diff
@@ -7,63 +7,139 @@ import {
...
-    users: PropTypes.array,
-    layoutOptions: PropTypes.object
+    layoutOptions: PropTypes.object,
+    collection: PropTypes.array,
+    pager: PropTypes.object
   };
 
+  constructor(props) {
+    super(props);
+    this.state = { limit: 20, page: 1 };
```
ayan4m1: These numbers could be defined as instance members like `this.defaultState = { limit: 20, page: 1};` and then you can re-use it in `componentDidMount()`

src/components/Dashboard/Users.js
```diff
@@ -7,63 +7,139 @@ import {
...
-    actions.requestUsers();
+    actions.requestUsers({ page: 1, limit: 20 });
+  }
+
+  pagerCounter() {
+    const { pages } = this.props.pager;
+
+    let i = 1;
+
+    const pager = [];
```
ayan4m1: I think part of the reason this PR is so large is because there is a lot of duplicated functionality in these components. You could define an [HOC](https://reactjs.org/docs/higher-order-components.html) called e.g. `withPagination` that adds these functions (pagerCounter, changePage, changeLimit, updateLimit) to the wrapped component, and I think it would cut a few thousand lines out of the PR.

src/components/Dashboard/Flavors.js
daviddyess: Updated

src/components/Dashboard/Flavors.js
daviddyess: Updated; same in changeLimit

src/components/Dashboard/Users.js
daviddyess: Updated in Pagination HOC

src/reducers/index.js
daviddyess: Removed

src/components/Dashboard/Home.js
```diff
@@ -0,0 +1,94 @@
...
+import { CardColumns, Card } from 'react-bootstrap';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+
+import { actions as dashboardActions } from 'reducers/dashboard';
+import { getStats } from 'selectors/dashboard';
+
+export class DashboardHome extends Component {
+  static propTypes = {
+    actions: PropTypes.object.isRequired,
+    stats: PropTypes.any
```
ayan4m1: Isn't this always an object?

daviddyess: Fixed

src/components/Dashboard/Main.js
```diff
@@ -0,0 +1,109 @@
...
+  Roles,
+  RoleAdd,
+  RoleEdit,
+  RoleDelete,
+  RoleUsers,
+  RoleAddUser,
+  RoleDeleteUser,
+  Users,
+  UserRoles
+} from 'components/Dashboard/';
```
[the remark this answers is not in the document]
daviddyess: Fixed

src/components/Dashboard/Flavors.test.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import { Provider } from 'react-redux';
+import renderer from 'react-test-renderer';
+import configureStore from 'redux-mock-store';
+
+import { initialState } from 'reducers/flavors';
+import ConnectedFlavors /* , { Flavors } */ from './Flavors';
```
ayan4m1: Commented out code

src/components/Dashboard/Role/DeleteUser.js
```diff
@@ -0,0 +1,87 @@
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import React, { Component } from 'react';
+import { bindActionCreators } from 'redux';
+import { Form as FinalForm } from 'react-final-form';
+import { Button, Form } from 'react-bootstrap';
+import {
+  DashboardLink as DashLink,
+  DashboardLayout as Layout
+} from 'components/Dashboard/';
```
ayan4m1: Trailing slash here

src/components/Dashboard/Role/Delete.js
```diff
@@ -0,0 +1,91 @@
...
+import { connect } from 'react-redux';
+import React, { Component } from 'react';
+import { bindActionCreators } from 'redux';
+import { Form as FinalForm } from 'react-final-form';
+import { Button, Form } from 'react-bootstrap';
+import {
+  DashboardLink as DashLink,
+  DashboardLayout as Layout,
+  RoleUsers
+} from 'components/Dashboard/';
```
ayan4m1: Trailing slash here

src/components/Dashboard/Role/Edit.js
```diff
@@ -0,0 +1,101 @@
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import React, { Component } from 'react';
+import { bindActionCreators } from 'redux';
+import { Form as FinalForm, Field } from 'react-final-form';
+import { Button, Form } from 'react-bootstrap';
+import {
+  DashboardLink as DashLink,
+  DashboardLayout as Layout
+} from 'components/Dashboard/';
```
ayan4m1: Trailing slash here

src/components/Dashboard/Role/AddUser.test.js
```diff
@@ -0,0 +1,88 @@
...
+        <RoutedRoleAddUser
+          actions={actions}
+          layoutOptions={defaultLayoutOptions}
+        />
+      </Provider>
+    );
+    const { instance } = component.root.findByType(RoleAddUser);
+
+    expect(instance).toBeDefined();
+    instance.handleSubmit(e);
```
ayan4m1: Why are there no assertions after this call?

src/components/Dashboard/Role/AddUser.test.js
```diff
@@ -0,0 +1,88 @@
...
+        <RoutedRoleAddUser
+          actions={actions}
+          layoutOptions={defaultLayoutOptions}
+        />
+      </Provider>
+    );
+    const { instance } = component.root.findByType(RoleAddUser);
+
+    expect(instance).toBeDefined();
+    instance.handleChange(e);
```
ayan4m1: Why no assertions after this call?

src/components/Dashboard/Role/AddUser.js
```diff
@@ -0,0 +1,139 @@
...
+            >
+              <span>Save</span>
+            </Button>
+          )}
+        </Form>
+        <br />
+        {isUserIdSelected && (
+          <UserRoles
+            layoutOptions={{ header: false, title: true, border: 'info' }}
+            userId={Number(userId)}
```
ayan4m1: Just curious, why does userId need to be Number()'d but roleId below does not?

src/components/Dashboard/Role/AddUser.js
```diff
@@ -0,0 +1,139 @@
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import React, { Component } from 'react';
+import { bindActionCreators } from 'redux';
+import { Button, Form } from 'react-bootstrap';
+import {
+  DashboardLink as DashLink,
+  DashboardLayout as Layout
+} from 'components/Dashboard/';
```
ayan4m1: Trailing slash here

src/components/Dashboard/Role/Add.js
```diff
@@ -0,0 +1,98 @@
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import React, { Component } from 'react';
+import { bindActionCreators } from 'redux';
+import { Form as FinalForm, Field } from 'react-final-form';
+import { Button, Form } from 'react-bootstrap';
+import {
+  DashboardLink as DashLink,
+  DashboardLayout as Layout
+} from 'components/Dashboard/';
```
ayan4m1: Trailing slash here

src/components/Dashboard/Roles.js
```diff
@@ -0,0 +1,105 @@
...
+    collection: PropTypes.array.isRequired,
+    pager: PropTypes.object.isRequired,
+    pagerNavigation: PropTypes.node.isRequired
+  };
+
+  render() {
+    const { collection, layoutOptions, pager, pagerNavigation } = this.props;
+    // Administrator and User roles aren't editable
+
+    let noEdit = false;
```
ayan4m1: Would you be able to refactor `noEdit` to a getter so that the assignment logic is not in the render method?

src/components/Dashboard/Roles.test.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import { Provider } from 'react-redux';
+import renderer from 'react-test-renderer';
+import configureStore from 'redux-mock-store';
+
+import { initialState } from 'reducers/roles';
+import { initialState as dashboardInitialState } from 'reducers/dashboard';
+import ConnectedRoles /* , { Roles } */ from './Roles';
```
ayan4m1: Commented out code

src/components/Dashboard/User/Roles.js
```diff
@@ -0,0 +1,117 @@
...
+            <thead>
+              <tr className="text-center">
+                <th>ID</th>
+                <th>Name</th>
+                <th>Options</th>
+              </tr>
+            </thead>
+            <tbody>
+              {roles.map((role, index) => {
+                style = {};
```
ayan4m1: Why not declare this variable const here?

src/components/Dashboard/User/Roles.js
```diff
@@ -0,0 +1,117 @@
...
+                if (roleId === role.Role.id) {
+                  style = { border: '2px solid red' };
```
ayan4m1: Is there a way we can do this without CSS in the JS?

src/components/Dashboard/Users.test.js
```diff
@@ -0,0 +1,55 @@
...
+    dashboard: dashboardInitialState
+  });
+
+  const RoutedUsers = withMemoryRouter(ConnectedUsers);
+
+  it('renders correctly', () => {
+    expect(
+      renderer
+        .create(
+          <Provider store={store}>
```
ayan4m1: Just a reminder for myself more than anything, we should have a withProvider that simplifies this boilerplate.

src/pages/index.js
```diff
@@ -1,6 +1,7 @@
 import { lazy } from 'react';
 
 export const imports = {
+  Dashboard: () => import(/* webpackChunkName: "recipe" */ 'pages/Dashboard'),
```
ayan4m1: I think this chunk name should be "admin" or "dashboard", rationale being that 99% of users who load the recipe chunk will never use this code - so it should be in its own chunk.

src/pages/index.test.js
```diff
@@ -11,6 +11,7 @@ import {
   NotFound,
   Profile,
   Recipe,
+  RecipeEditor,
```
ayan4m1: Oh, looks like I forgot to add this test here...

src/sagas/dashboard.js
```diff
@@ -0,0 +1,113 @@
+import { all, call, put, takeLatest } from 'redux-saga/effects';
+
+import request from 'utils/request';
+import { actions, types } from 'reducers/dashboard';
+import { actions as toastActions } from 'reducers/toast';
+
+function* requestDashboardWorker(dashboardComponent) {
+  yield put(actions.requestDashboardSuccess(dashboardComponent));
+}
```
ayan4m1: Mind adding newlines between these funcs?

src/components/Pagination/Pagination.test.js
```diff
@@ -0,0 +1,104 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+// import withPagination, { pagination, mapStateToProps, mapDispatchToProps } from './Pagination';
```
ayan4m1: Commented out code?

src/components/Pagination/Pagination.test.js
```diff
@@ -0,0 +1,104 @@
...
+      name: 'Anise',
+      slug: null,
+      density: null,
+      Vendor: { id: 3, name: 'Capella', slug: 'capella', code: 'CAP' }
+    }
+  ];
+  const pager = { count: 861, limit: 5, page: 1, pages: 173 };
+  const TestComponent = () => <p />;
+  const Component = pagination(TestComponent);
+  // const ConnectedComponent = withPagination(TestComponent);
```
ayan4m1: Commented out code

src/components/Dashboard/Role/AddUser.js
daviddyess: It's because userId is from the select field, which sets it as a string, leading to a validation warning in UserRoles for being a string. I moved this to the handleChange to hopefully make it more obvious why it's needed.

src/components/Dashboard/Roles.js
daviddyess: Easy fix

src/components/Dashboard/Role/AddUser.test.js
daviddyess: Fixed

src/components/Dashboard/Role/AddUser.test.js
daviddyess: Fixed

PR #44, opened 2019-09-04
Header User Conditionals / Additionals

daviddyess:
- Header conditionals:
  - Hide Login/Register if user is logged in
  - Display User if user is logged in
- Added Dashboard link to User (will require a role check in the future)
- Added Logout link to User


src/components/Header/Header.js
```diff
@@ -48,22 +71,41 @@ export default class Header extends Component {
...
+                        onClick={this.logoutUser}
+                      >
+                        Logout
+                        <br />
+                      </NavDropdown.Item>
+                    </NavDropdown>
+                  ) : (
+                    ''
+                  )}
+                  {!loggedIn ? this.renderNavItem('/login', 'Login') : ''}
```
ayan4m1: Might want to wrap these two in a `<Fragment>` so you only need one ternary.

src/sagas/application.js
```diff
@@ -172,6 +172,39 @@ function* loginUserWorker({ emailAddress, password }) {
   }
 }
 
+function* logoutUserWorker() {
+  try {
+    // first, check to see if we are logged in
+    const loggedIn = yield select(isLoggedIn);
+
+    if (loggedIn) {
+      localStorage.setItem('accessToken', null);
```
ayan4m1: Probably want `localStorage.removeItem` here - [docs](https://developer.mozilla.org/en-US/docs/Web/API/Storage/removeItem)

src/sagas/application.js
```diff
@@ -172,6 +172,39 @@ function* loginUserWorker({ emailAddress, password }) {
...
+      localStorage.setItem('expiration', null);
+    }
+
```
ayan4m1: Don't we need to clear the accessToken and expiration from redux state as well?

src/components/Header/Header.js
daviddyess: This has been implemented

src/sagas/application.js
daviddyess: logoutUserSuccess resets the 'authorization' state to initialState

src/sagas/application.js
daviddyess: Refactored to `localStorage.removeItem()`.  I had read somewhere that `localStorage.setItem()` is better for performance than deleting and creating, but I can't find the reference; in this case I doubt there it makes a difference one way or the other.

src/components/Header/Header.js
```diff
@@ -104,8 +104,14 @@ export class Header extends Component {
...
                   )}
-                  {!loggedIn ? this.renderNavItem('/login', 'Login') : ''}
-                  {!loggedIn ? this.renderNavItem('/register', 'Register') : ''}
+                  {!loggedIn ? (
+                    <Fragment>
+                      {this.renderNavItem('/login', 'Login')}
+                      {this.renderNavItem('/register', 'Register')}
+                    </Fragment>
+                  ) : (
+                    ''
```
ayan4m1: Can this be `null` instead of an empty string?

daviddyess: Updated

daviddyess:
@ayan4m1 any other change requests for this PR?

src/components/Header/Header.test.js
```diff
@@ -1,18 +1,69 @@
...
+  it('renders logged out correctly', () => {
+    const props = {
+      loggedIn: false
+    };
+
+    expect(
+      renderer
+        .create(
+          <Provider store={store}>
+            <RoutedHeader actions={actions} {...props} />
```
ayan4m1: If this is the inner, unconnected Header export, then there should not be a need for a `<Provider>` - if `Header` itself uses/references a connected component, that can be mocked like

```
jest.mock('components/SomeComponent/SomeComponent', () =>
  require('utils').mockComponent('SomeComponent')
);
```

src/components/Header/__snapshots__/Header.test.js.snap
```diff
@@ -170,8 +170,127 @@ exports[`<Header /> renders correctly 1`] = `
...
+              aria-current={null}
+              className="nav--link-custom px-3 nav-link"
+              data-rb-event-key={null}
+              disabled={false}
+              href="/calculator"
+              onClick={[Function]}
+              role="tab"
+              tabIndex={-1}
+            >
+              Calculator
```
ayan4m1: Why is the logged-out snapshot showing items that shouldn't render for it?

daviddyess: I added a check for Calculator, are there any other items that don't need to be displayed?

src/components/Header/Header.test.js
```diff
@@ -15,50 +11,33 @@ describe('<Header />', () => {
...
 
     expect(
-      renderer
-        .create(
-          <Provider store={store}>
-            <RoutedHeader actions={actions} {...props} />
-          </Provider>
-        )
-        .toJSON()
+      renderer.create(<RoutedHeader {...props} />).toJSON()
```
ayan4m1: 👏

PR #61, opened 2020-02-15
Add footer

ayan4m1:
This PR adds a trivial/ugly footer to each page. [It's just a jumping-off point](https://y.yarn.co/a26767df-b84c-4ede-9cde-fba6174d58d3_screenshot.jpg)


daviddyess:
Looks like you just need to refresh your snapshot? Everything else looks like a good "jumping-off point" ;)

PR #65, opened 2020-02-17
Saga utilities

daviddyess:
- Added functions to assist in fetching paged data sets (`counter()`, `pager()`, `paged()`)
- Added function to assist in creating toasts (`toast()`, `errorToast()`, `successToast()`)
- Added a `helper` object to import all of the functions to be usable as ie. `helper.paged()`
- Added comments to show examples of usage


src/utils/saga/index.js
```diff
@@ -0,0 +1,11 @@
+import { counter, pager, paged } from './paged';
+import { errorToast, successToast, toast } from './toasts';
+
+export const helpers = {
```
ayan4m1: This could be

```js
import pageUtilities from './paged';
import toastUtilities from './toasts';

export default {
  ...pageUtilities,
  ...toastUtilities
}
```

Then you can still `import helpers from 'utils/saga';` or do a "destructured" import `import { pager } from 'utils/saga';` - but you can't destructure in an import so you'd never be able to do something like `import { helpers: { pager } } from 'utils/saga';` with the current implementation.

src/utils/saga/paged.js
```diff
@@ -0,0 +1,123 @@
...
+    store: rolesPager
+  },
+  path: {
+    counter: '/roles/count',
+    query: '/roles/'
+  },
+  type: 'Roles' }
+ */
+export function* pager({ req }) {
+  const Pager = {};
```
ayan4m1: This function should return a class instance rather than building up an object with properties.

src/utils/saga/paged.js
```diff
@@ -0,0 +1,123 @@
...
+    (count > Number(limit) && cached[page].length !== Number(limit)) ||
+    (count < Number(limit) && cached[page].length !== count)
+  ) {
+    let offset = page * limit - limit + 1;
+
+    if (offset > count) {
+      offset = count - limit;
+    }
+    const endpoint = {
+      url: `${path.query}?limit=${limit}&offset=${offset}`,
```
ayan4m1: Will we ever need to _append_ to an existing query string? i.e. if `path.query` is already `/some/call?withparams=likethis`

src/utils/saga/toasts.js
```diff
@@ -0,0 +1,55 @@
...
+ * @example
+ * { title: 'Toast Title' (optional),
+ *   icon: 'times-square' (optional),
+ *   message: 'My response message' (required) }
+ */
+export function* toast(msg) {
+  const { icon, title, message } = msg;
+  const response = {};
+
+  response.title = title ? title : 'Server Message';
```
ayan4m1: Why not initialize these inline with the object above?

src/utils/saga/paged.js
daviddyess: I needed to remove confusion here, so I renamed path to route and path.query to route.data. It is just the route to call (/roles for example), the parameters are appended after the pager determines offset.

src/utils/saga/index.js
daviddyess: Updated

src/utils/saga/paged.js
daviddyess: I restructured this, so pager and paged are just a single function - pager. I couldn't think of an instance when they were would needed separately.

src/utils/saga/toasts.js
daviddyess: fixed

src/utils/saga/counter.test.js
```diff
@@ -0,0 +1,41 @@
+import { call } from 'redux-saga/effects';
+import request from 'utils/request';
+import { counter } from './counter';
+
+/** Example Usage
+const response = yield call(helpers.counter, { url: '/roles/count', type: 'Roles' });
+ */
+
+describe('pager saga helpers', () => {
```
ayan4m1: Should be `counter saga helpers` right?

src/utils/saga/pager.js
```diff
@@ -0,0 +1,77 @@
...
+    pager: { store },
+    route,
+    type
+  } = req;
+
+  const Pager = {
+    count: !store.count
+      ? yield call(counter, { url: route.count, type })
+      : store.count,
+    limit: !req.pager.limit ? store.limit : req.pager.limit,
```
ayan4m1: This can be `req.pager.limit || store.limit`, right?

src/utils/saga/pager.js
```diff
@@ -0,0 +1,77 @@
...
+    page: !req.pager.page ? store.page : req.pager.page
+  };
+
+  Pager.pages =
+    !store.pages || store.pages === null || req.pager.limit !== store.limit
```
ayan4m1: You can use optional chaining here

src/utils/saga/counter.test.js
daviddyess: fixed

src/utils/saga/pager.js
daviddyess: fixed

src/utils/saga/pager.js
```diff
@@ -28,7 +28,7 @@ export function* pager(req) {
     count: !store.count
       ? yield call(counter, { url: route.count, type })
       : store.count,
-    limit: !req.pager.limit ? store.limit : req.pager.limit,
+    limit: req.pager.limit || store.limit,
     page: !req.pager.page ? store.page : req.pager.page
```
ayan4m1: Oh, sorry I should have mentioned, same thing here, `||`

PR #68, opened 2020-03-08
Initial Flavors Page w/ Flavor Stash Overlay

daviddyess:
Beginning of the Flavors page, provides a means to add/remove flavors to/from the user's Flavor Stash. Adds testing for Flavor redux. Table display is considered interim and to be refactored (ie. Flavor links, recipes, abbreviations, etc).

Updated ToggleButton component to allow parent component to control value (icon), title.


src/pages/Flavors.js
```diff
@@ -1,36 +1,198 @@
...
+
+  constructor(props) {
+    super(props);
+
+    this.state = { stashControl: false };
+
+    const { actions, loggedIn, stashLoaded } = this.props;
+
+    if (loggedIn) {
+      if (!stashLoaded) {
```
ayan4m1: Actions are usually called in componentDidMount - not sure it makes an actual difference here but there's no point in requesting the stash until the component is actually going to be shown imo.

src/pages/Flavors.js
```diff
@@ -1,36 +1,198 @@
...
+        actions.requestStash();
+      }
+      this.handleStashControl = this.stashController.bind(this);
+      this.handleAddToStash = this.addToStash.bind(this);
+      this.handleRemoveFromStash = this.removeFromStash.bind(this);
+    }
+  }
+
+  stashController(event) {
```
ayan4m1: This seems like it should have a verb at the end of it, as it's being invoked in order to change state.

src/pages/Flavors.js
```diff
@@ -1,36 +1,198 @@
...
+    const { stash } = this.props;
+    const { holdings: stashMap } = this.state;
+
+    if (!stashMap) {
+      const holdings = [];
+
+      stash.map(flavor => {
+        holdings[flavor.flavorId] = true;
```
ayan4m1: Holdings should be an object here - otherwise if you had a flavor with ID 10001 you'd be allocating 10000 empty array entries before it in order to store everything.

src/pages/Flavors.js
```diff
@@ -1,36 +1,198 @@
...
+      });
+      this.setState({ holdings });
```
ayan4m1: See if there is a way to call this.setState once during this method invocation - it will help keep complexity of tracking state changes down as the component grows.

src/pages/Flavors.js
```diff
@@ -1,36 +1,198 @@
...
+    actions.removeStash({ id });
+
+    const { holdings } = this.state;
+
+    holdings[id] = false;
+
+    this.setState({ holdings });
+  }
+
+  inStashIcon(id) {
```
ayan4m1: inStashIcon and noStashIcon should be the same getter - `onClick` would then have an `if { } else { }` or ternary construct that invokes the proper method based on the component state. We should change ToggleButton to pass the current state as a boolean instead of `event` to the `onClick` handler here, that will make this much easier.

src/pages/Flavors.js
```diff
@@ -1,36 +1,198 @@
...
+          </Row>
+          <Row>
+            <Col className="text-center">
+              {loggedIn ? (
+                <Form>
+                  <Form.Check
+                    name="stashControl"
+                    type="checkbox"
+                    id="flavorStash"
+                    label="&nbsp;Enable Flavor Stash"
```
ayan4m1: Maybe outside the scope of this PR, but we shouldn't have to prepend &nbsp; in here to get the appropriate padding on this control - @JosefBud any ideas?

src/pages/Flavors.test.js
```diff
@@ -0,0 +1,251 @@
...
+import { Provider } from 'react-redux';
+import renderer from 'react-test-renderer';
+import configureStore from 'redux-mock-store';
+import { initialState as appInitialState } from 'reducers/application';
+import { initialState as flavorInitialState } from 'reducers/flavor';
+import { initialState as flavorsInitialState } from 'reducers/flavors';
+import ConnectedFlavors, { Flavors } from './Flavors';
+import { withMemoryRouter } from 'utils';
+
+describe('Dashboard <Flavors />', () => {
```
ayan4m1: This isn't a Dashboard component, right?

src/pages/Flavors.test.js
```diff
@@ -0,0 +1,251 @@
...
+    flavor: flavorInitialState,
+    flavors: flavorsInitialState
+  });
+
+  const notLoggedInProps = {
+    actions,
+    collection,
+    loggedIn: false,
+    pager,
+    pagerNavigation: <p></p>,
```
ayan4m1: This could be declared as a const to avoid having to redefine it.

src/reducers/flavor.js
```diff
@@ -3,7 +3,11 @@ import { buildActions } from 'utils';
 export const types = buildActions('flavor', [
   'REQUEST_STASH',
   'REQUEST_STASH_SUCCESS',
-  'REQUEST_STASH_FAILURE'
+  'ADD_TO_STASH',
```
ayan4m1: I think this should just be `ADD_STASH` - thoughts?

src/reducers/flavor.js
```diff
@@ -40,7 +67,12 @@ export const reducer = (state = initialState, action = {}) => {
         stash: action.stash,
         loaded: true
       };
-    case types.REQUEST_STASH_FAILURE:
+    case types.UPDATE_STASH_SUCCESS:
+      return {
+        ...state,
+        loaded: false
+      };
+    case types.REQUEST_FAILURE:
```
ayan4m1: I think this is fine, but can you think of any instance where we'd need to know _which_ request failed?

src/sagas/flavor.js
```diff
@@ -1,27 +1,25 @@
...
-      yield put(appActions.requestCurrentUser());
-      yield take([
-        appTypes.REQUEST_CURRENT_USER_SUCCESS,
-        appTypes.REQUEST_CURRENT_USER_FAILURE
-      ]);
-      user = yield select(getUser);
+    const loaded = yield select(isLoaded);
+
+    if (loaded) {
+      return true;
```
ayan4m1: I think this can just `return;`

src/sagas/flavor.js
```diff
@@ -38,22 +36,140 @@ function* requestStashWorker() {
       throw new Error('Request failed for an unspecified reason!');
     }
   } catch (error) {
-    // eslint-disable-next-line
-    console.dir(error);
```
ayan4m1: 💯

src/sagas/flavor.js
```diff
@@ -38,22 +36,140 @@ function* requestStashWorker() {
...
-    yield put(actions.requestStashFailure(error));
+    yield put(actions.requestFailure(error));
+  }
+}
+
+function* addStashWorker({ flavor }) {
+  try {
+    const user = yield call(getCurrentUser);
+
+    const endpoint = {
```
ayan4m1: Should have a check for null user or a `user?.id` to prevent a ReferenceError

src/sagas/flavor.js
```diff
@@ -38,22 +36,140 @@ function* requestStashWorker() {
...
+      url: `/user/${user.id}/flavor/${flavor.id}`,
+      method: 'DELETE'
+    };
+
+    const result = yield call(request.execute, { endpoint });
+
+    if (result.success) {
+      yield put(actions.updateStashSuccess());
```
ayan4m1: Is there any need for a `removeStashSuccess`?

src/reducers/flavor.js
daviddyess: This was more of a trial, but there is/was only one `error` property. Maybe I'm not seeing the bigger picture, but there was nothing to distinguish which request failed previously.

src/sagas/flavor.js
daviddyess: I had one originally, to remove any residual artifacts. Currently that is only the flavor note, if I'm not mistaken, and we should probably preserve those.

src/pages/Flavors.js
daviddyess: Moved to componentDidMount

src/pages/Flavors.js
daviddyess: Renamed to handleStashToggle

src/pages/Flavors.js
daviddyess: holdings = {}

src/pages/Flavors.js
daviddyess: Doneskis

src/pages/Flavors.js
daviddyess: Merged

src/reducers/flavor.js
daviddyess: Renamed

src/sagas/flavor.js
daviddyess: getCurrentUser() throws an error if user is not truthy. I added a check for user.id though, just in case.

src/sagas/flavor.js
daviddyess: I had it return true, because I didn't like the test expecting `undefined`.  lol Removed it

src/pages/Flavors.js
JosefBud: I'm at work and unable to actually view these changes in a build, so forgive me if I'm speaking out of ignorance, but shouldn't we be using the new <ToggleButton /> component rather than a checkbox if it's outside of a form?

daviddyess:
It's in a form, I just thought it looked confusing having the same button
as add/remove. I think a different on/off toggle would be more appropriate,
personally.

JosefBud:
> It's in a form, I just thought it looked confusing having the same button
as add/remove. I think a different on/off toggle would be more appropriate,
personally.
@daviddyess 

Might I suggest using the toggle switch I've got going in the `design/recipe-search` branch? It's just a fancy styled checkbox.
https://github.com/mixnjuice/frontend/blob/c5519bfa9d754d55dfe9c52e96f078b2dd7e6d1d/src/pages/Recipes.scss#L65-L154
https://github.com/mixnjuice/frontend/blob/c5519bfa9d754d55dfe9c52e96f078b2dd7e6d1d/src/pages/Recipes.js#L451-L470

src/components/ToggleButton/ToggleButton.js
```diff
@@ -13,36 +13,38 @@ export default class ToggleButton extends Component {
...
+    buttonProps: {},
+    iconProps: {},
+    value: false,
     variant: 'check'
   };
 
   constructor(props) {
     super(props);
 
-    this.state = { value: this.props.initialValue };
```
ayan4m1: 💯

src/pages/Flavors.js
ayan4m1: Yeah, good point, can you think of an icon pairing for show/hide? I know I need to add an expand/contract for the Recipe Editor.

src/sagas/flavor.js
```diff
@@ -1,27 +1,29 @@
...
-        appTypes.REQUEST_CURRENT_USER_FAILURE
-      ]);
-      user = yield select(getUser);
+    if (loaded) {
+      return;
+    }
+
+    const user = yield call(getCurrentUser);
+
+    if (!user.id) {
```
ayan4m1: This should probably be `user?.id` in case user is null

src/components/ToggleButton/ToggleButton.js
```diff
@@ -5,64 +5,87 @@ import { Button } from 'react-bootstrap';
...
-    const classes = classNames(className, 'btn-toggle');
+    if (iconOnly) {
+      const classes = classNames(className, 'icon-toggle');
+
+      return (
+        <FontAwesomeIcon
+          size="lg"
+          {...iconProps}
+          icon={icon}
+          {...buttonProps}
```
ayan4m1: This might be unintuitive to a developer unfamiliar with the implementation, is there a need for this spreading to occur?

src/pages/Flavors.js
```diff
@@ -48,7 +48,7 @@ export class Flavors extends Component {
     }
   }
 
-  handleStashToggle(event) {
+  /* handleStashToggle(event) {
```
ayan4m1: Commented code?

src/pages/Flavors.js
```diff
@@ -66,6 +66,40 @@ export class Flavors extends Component {
...
+        holdings[flavor.flavorId] = true;
+      });
+
+      this.setState({ stashToggle: !stashToggle, holdings });
+    } else {
+      this.setState({ stashToggle: !stashToggle });
+    }
+  }
+
+  stashToggle() {
```
ayan4m1: I think this should be a getter.

src/pages/Flavors.js
```diff
@@ -66,6 +66,40 @@ export class Flavors extends Component {
...
+    const { stashToggle } = this.state;
+
+    return (
+      <ToggleButton
+        value={stashToggle}
+        onClick={e => this.handleStashToggle(e)}
+        buttonProps={{
+          title: stashToggle ? 'Enable Flavor Stash' : 'Disable Flavor Stash'
```
ayan4m1: I think this might be backwards - if stashToggle is true then clicking the button would disable it, right?

PR #69, opened 2020-03-15
<ToggleButton> changes

JosefBud:
- Added a checkbox styled to look like a modern "toggle" switch
- Added two variants - "switch" and "grid-list"
  - Switch: When in an "off" state, background is grey. When in an "on" state, background is teal. Toggle is rendered alone with no icons or text around it.
  - Grid-List: Background is teal regardless of state, and has an icon on each side: a grid icon and a list icon
- Updated tests


src/components/ToggleButton/ToggleButton.js
```diff
@@ -55,14 +71,28 @@ export default class ToggleButton extends Component {
...
         icon = value ? ['fas', 'check-square'] : ['far', 'square'];
         break;
     }
+
+    if (variant === 'switch') {
+      return this.renderSwitch(className, value, title);
+    }
+
+    if (variant === 'grid-list') {
+      const classes = classNames(className, 'slider-teal');
```
ayan4m1: Could this be `return this.renderGridList();`?

src/components/ToggleButton/ToggleButton.scss
```diff
@@ -6,3 +6,111 @@
...
+    --webkit-transition: 0.4s;
+    transition: none;
+
+    &.slider-teal {
+      background-color: $teal;
+    }
+  }
+}
+
+.slider {
```
ayan4m1: You could refactor this to a mixin in order to remove the duplication here and have something like

```scss
@media screen and (prefers-reduced-motion: reduce) {
  .slider {
    @include with-slider(false);
  }
}

.slider {
  @include with-slider(true);
}
```

src/components/ToggleButton/ToggleButton.scss
```diff
@@ -6,3 +6,111 @@
...
+}
+
+input:checked + .slider::before {
+  --webkit-transform: translateX(26px);
+  --ms-transform: translateX(26px);
+  transform: translateX(26px);
+}
+
+/* Rounded sliders */
+.slider.round {
```
ayan4m1: To be more BEM, I think this should be `.slider--round`

src/components/ToggleButton/ToggleButton.scss
```diff
@@ -6,3 +6,111 @@
...
+  border-radius: 34px;
+}
+
+.slider.round::before {
+  border-radius: 50%;
+}
+
+.grid-list {
```
ayan4m1: Maybe .slider--grid-list?

src/components/ToggleButton/ToggleButton.scss
JosefBud: The duplication was caused by stylelint, I believe. I cleaned it up but without using mixins. If you still want me to turn the slider into a mixin I can do that, just let me know!

Check it out here: https://github.com/mixnjuice/frontend/pull/69/commits/9913d051705c3e5053ed90c93d45523e9141632a

src/components/ToggleButton/ToggleButton.js
```diff
@@ -57,6 +57,24 @@ export default class ToggleButton extends Component {
...
   }
 
+  renderGridList(classes, value, title) {
+    const joinedClasses = classNames(classes, 'slider-teal');
+
+    return (
+      <div className="slider--grid-list">
+        <FontAwesomeIcon
+          icon={['fas', 'list']}
+          className="slider--grid-list--icon"
```
ayan4m1: This should be `slider-icon--grid-list` - so there is `slider` and its modifiers, `slider-icon` and its modifiers

src/components/ToggleButton/ToggleButton.js
```diff
@@ -82,15 +100,7 @@ export default class ToggleButton extends Component {
     }
 
     if (variant === 'grid-list') {
```
ayan4m1: Also, sorry I didn't see this at first, but I'm thinking we should just have a switch statement for variant at this point, where `default:` handles all the non-special-case variants.

src/components/ToggleButton/ToggleButton.scss
ayan4m1: Nah, that's fine, looks good to me!

src/index.html
```diff
@@ -2,6 +2,7 @@
 <html>
   <head>
     <title>MixNJuice</title>
+    <link rel="icon" type="image/png" href="/media/favicon.png" />
```
ayan4m1: 🎺

src/pages/__snapshots__/Flavors.test.js.snap
```diff
@@ -182,9 +182,19 @@ exports[`Page <Flavors /> renders correctly when stash is not loaded 1`] = `
       <div
         className="text-left ml-2 mb-4 col"
       >
-        <i
-          className="fa fas-toggle-off"
-        />
+        <label
```
ayan4m1: Would you mind changing Flavors.js to add a line mocking the ToggleButton component? That will simplify this part of the snapshot and avoid it breaking if the component logic changes again.

src/pages/__snapshots__/Flavors.test.js.snap
```diff
@@ -182,19 +182,7 @@ exports[`Page <Flavors /> renders correctly when stash is not loaded 1`] = `
...
-            checked={false}
-            name="Enable Flavor Stash"
-            onChange={[Function]}
-            type="checkbox"
-          />
-          <span
-            className="slider slider--round"
-          />
-        </label>
+        <ToggleButton />
```
ayan4m1: 💯

PR #70, opened 2020-03-19
Dark mode

hylophile:
Adds a ToggleButton to user settings to toggle Dark Mode.
Can easily be refactored for more themes.

Still to do:
- complete the dark mode theme (mostly done)
- add persistency of the setting (save in db or localStorage?)


src/_color-themes.scss
```diff
@@ -0,0 +1,103 @@
+// Themes
+
+$bg: "bg";
+$fg: "fg";
+$lightest-blue: "lightest-blue";
```
ayan4m1: We should think about refactoring all of these to a $colorNames map in variables.scss

src/_color-themes.scss
```diff
@@ -0,0 +1,103 @@
...
+$dark-gray: "dark-gray";
+$buttonborder-custom: "buttonborder-custom";
+$form-control-border: "form-control-border";
+$input-group-text-border: "input-group-text-border";
+
+$themes: (
+  default: (
+    bg: #fff,
+    fg: gray("800"),
+    lightest-blue: #e7f0ff,
```
ayan4m1: Are these duplicates of the values in variables.scss? If so, I think we should re-use what is declared there somehow. If not, I am wondering where they came from.

src/_color-themes.scss
```diff
@@ -0,0 +1,103 @@
+// Themes
+
+$bg: "bg";
+$fg: "fg";
+$lightest-blue: "lightest-blue";
+$lighter-blue: "lighter-blue";
+$light-blue: "light-blue";
+$blue-custom: "blue-custom";
```
ayan4m1: What is the `-custom` suffix for here?

src/_variables.scss
```diff
@@ -1,12 +1,14 @@
-$lightest-blue: #e7f0ff;
-$lighter-blue: #d3e3fc;
-$light-blue: #b1cefd;
-$blue: #77a6f7;
-$teal: #00887a;
-$baby-blue: #00bad3;
-$light-red: #fd8383;
-$lighter-red: #ffccbc;
-$shadow-grey: #777;
+// $lightest-blue: #e7f0ff;
```
ayan4m1: Why are these commented out?

src/_color-themes.scss
hylophile: Another option would be to omit these. I only added them to be able to write `t($bg)` instead of `t("bg")`, so it's merely a matter of aesthetics which doesn't make a big difference.

src/_color-themes.scss
hylophile: Yes they are and I agree. In a similar vein, I think it's better to use variable names like `$main-accent` instead of `$teal` in the themes. The current state is quite confusing.

src/_color-themes.scss
hylophile: I think what happened was that the mapping got confused by names like `$teal`,`$black` etc. because they are standard colors and the theming wouldn't apply. When applying the renaming mentioned before this would become unnecessary though.

src/_variables.scss
hylophile: I did this for debugging purposes, but it will be reverted.

src/_color-themes.scss
ayan4m1: Yeah, I don't think it's necessary to have them as variables since they're only used in this file.

src/_color-themes.scss
ayan4m1: @JosefBud You OK with changing the var names to be semantic rather than literal?

src/_color-themes.scss
```diff
@@ -0,0 +1,120 @@
...
+      }
+      /* stylelint-enable */
+      @content;
+      $theme-map: null !global;
+    }
+  }
+}
+
+// body rule needs special treatment, although this could probably be heavily simplified
+@mixin themed_body {
```
ayan4m1: If we applied these styles to a container instead of the `body` tag itself, could we avoid having this duplication?

src/_color-themes.scss
```diff
@@ -0,0 +1,120 @@
...
+        ) !global;
+      }
+      /* stylelint-enable */
+      @content;
+      $theme-map: null !global;
+    }
+  }
+}
+
+@function t($key) {
```
ayan4m1: Do you mind renaming this to `theme`? I feel like it's going to be pretty confusing to newcomers otherwise.

src/_variables.scss
```diff
@@ -1,12 +1,22 @@
...
+$blue-themed: #77a6f7;
 $teal: #00887a;
-$baby-blue: #00bad3;
-$light-red: #fd8383;
-$lighter-red: #ffccbc;
-$shadow-grey: #777;
+$teal-dark: #009483;
+$blue-baby: #00a5bb;
+$red-light: #fd8383;
+$red-lighter: #ffccbc;
```
ayan4m1: These shades should probably go into a map like

```scss
$red: (
  "normal": #ff0000,
  "lighter": #ffccbc;
  "light": #fd8383
)
```

src/components/App/App.test.js
```diff
@@ -55,7 +58,8 @@ describe('<App />', () => {
       user: {
         id: 123,
         name: 'Doe'
-      }
+      },
+      theme: { name: 'default' }
```
ayan4m1: If the component is mocked above, why is this prop needed?

src/components/Dashboard/__snapshots__/Users.test.js.snap
```diff
@@ -38,18 +38,25 @@ exports[`Dashboard <Users /> renders correctly 1`] = `
         <div
           className="text-right col"
         >
-          <select
-            onBlur={[Function]}
-            onChange={[Function]}
-            value={1}
+          <div
```
ayan4m1: Why did all these snapshots change if no code changed with them?

src/components/Pagination/Pagination.js
```diff
@@ -81,17 +81,20 @@ export const pagination = WrappedComponent =>
...
-                onChange={this.handlePageChange}
-                onBlur={this.handlePageChange}
-              >
-                {this.pagerCounter().map((value, i) => (
-                  <option value={value} key={i}>
-                    Page {value}
-                  </option>
-                ))}
-              </select>
+              <div className="select">
```
ayan4m1: I actually have a PR out to replace most of this implementation, would you mind reverting these changes?

src/components/Pagination/Pagination.js
```diff
@@ -81,17 +81,20 @@ export const pagination = WrappedComponent =>
...
+                <span className="arr"></span>
```
ayan4m1: Why is this being added?

src/components/ThemeContainer/ThemeContainer.js
```diff
@@ -0,0 +1,38 @@
...
+import PropTypes from 'prop-types';
+import { getThemeName } from 'selectors/theme';
+import { connect } from 'react-redux';
+
+export class ThemeContainer extends Component {
+  static propTypes = {
+    theme: PropTypes.string.isRequired
+  };
+
+  constructor(props) {
```
ayan4m1: No need for the constructor if all it does is call super.

src/components/ThemeContainer/ThemeContainer.js
```diff
@@ -0,0 +1,38 @@
...
+      document.body.classList.add('theme--default');
+      document.body.classList.remove('theme--dark');
+    } else {
+      document.body.classList.add('theme--dark');
+      document.body.classList.remove('theme--default');
+    }
+  }
+
+  render() {
+    return null;
```
ayan4m1: Any time a component returns null, it's a good sign that it should be a [higher-order component](https://reactjs.org/docs/higher-order-components.html) instead - it doesn't need to exist in the DOM tree.

src/components/ThemeContainer/__snapshots__/ThemeContainer.test.js.snap
```diff
@@ -0,0 +1,5 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`<ThemeContainer /> renders connected component correctly 1`] = `null`;
```
ayan4m1: If this were a HOC, then these snapshots could go away, as they don't really serve a purpose right now.

src/components/ToastDrawer/ToastDrawer.scss
```diff
@@ -4,3 +4,24 @@
   z-index: 100;
   padding-right: 40px;
 }
+
+.toast-header {
+  @include themed {
+    background-color: rgba(t("toast-bg"), 0.85);
```
ayan4m1: @JosefBud Thoughts on partially transparent toasts? I'm not sure I like the idea of losing contrast ratio just for the sake of being able to see behind it.

src/index.scss
```diff
@@ -4,6 +4,9 @@
 // our variables
 @import "./variables";
 
+// color themes
+@import "./color-themes";
```
ayan4m1: Do you mind renaming this file to `themes` ?

src/pages/user/Settings.js
```diff
@@ -1,8 +1,19 @@
...
+import { actions as themeActions } from 'reducers/theme';
+import { getThemeName } from 'selectors/theme';
+import { connect } from 'react-redux';
+import { bindActionCreators } from 'redux';
+import ToggleButton from 'components/ToggleButton/ToggleButton';
+import PropTypes from 'prop-types';
+
+export class UserSettings extends Component {
+  static propTypes = {
+    theme: PropTypes.string,
```
ayan4m1: Do you mind sorting the props alphabetically, at least for the propTypes declaration?

src/pages/user/Settings.js
```diff
@@ -106,6 +125,22 @@ export default class UserSettings extends Component {
               </InputGroup>
             </Col>
           </Row>
+          <Row></Row>
```
ayan4m1: Why is this empty Row here?

src/pages/user/Settings.js
```diff
@@ -33,6 +44,14 @@ export default class UserSettings extends Component {
     this.setState({ filename: event.target.files[0].name });
   }
 
+  toggleDarkMode() {
```
ayan4m1: This logic should be in the new theme reducer - you should be able to call `this.props.actions.toggleDarkMode()` and be done with it, imo.

src/reducers/theme.test.js
```diff
@@ -0,0 +1,20 @@
...
+      type: types.SET_THEME,
+      theme: name
+    });
+  });
+
+  it('reduces SET_THEME action', () => {
+    const action = actions.setTheme(name);
+
+    expect(reducer({}, action)).toEqual({
+      name: 'dracula'
```
ayan4m1: You can just do `.toEqual({ name })` here

src/components/Pagination/Pagination.js
hylophile: As soon as the `select` element is styled, it loses its integrity so to speak. The arrow on the right side looks completely different, for example. Also, it seems the select element itself doesn't seem to be stylable nicely because its looks are decided by the browser if we don't override them. This is why the select class (for the button) and the arr class (the arrow on the right side) were added here. As a bonus, it looks more fitting to the whole theme.

src/components/ToastDrawer/ToastDrawer.scss
hylophile: Note I tried to stay as close as possible to the current styling rules as possible, these were just implicit rules.

src/pages/user/Settings.js
hylophile: I wanted to add some space in between the togglebutton and the content above, but there's probably better ways to do that.

src/_color-themes.scss
hylophile: When I did that, the problem was that when the app didn't span the whole page, there was an unstyled part beneath it.

src/_variables.scss
hylophile: Done - the only problem is `$teal`, I'm assuming as it is a standard color name it's not possible to have a `$teal` map. I don't understand why it works for `$red` though.

src/components/App/App.test.js
hylophile: Removed.

src/components/Pagination/Pagination.js
hylophile: Done. Hope to reuse the select element at some point though.

src/components/ThemeContainer/ThemeContainer.js
hylophile: Removed.

src/pages/user/Settings.js
hylophile: Added toggleDarkMode() action and removed setTheme() action

src/_variables.scss
```diff
@@ -1,12 +1,33 @@
...
+  lighter: #d3e3fc,
+  lightest: #e7f0ff,
+  normal: #77a6f7
+);
+
+$gray: (
+  200: #e9ecef,
+  300: #dee2e6,
+  800: #343a40,
+  dark: #333,
```
hylophile: Maybe these should be renamed to lightest, lighter etc. to be more clear? Also, `800` and `dark` are very similar, they could be merged into one.

src/themes.scss
```diff
@@ -0,0 +1,120 @@
...
+    toast-bg: $white-pure,
+    toast-close: $black-pure
+  ),
+  dark: (
+    bg: map-get($gray, darkest),
+    fg: map-get($gray, 200),
+    alt-accent-lightest: #121925,
+    alt-accent-lighter: #0f1e38,
+    alt-accent-light: #1761da,
+    alt-accent: #1959c9,
```
hylophile: I wasn't sure if I should add variables / include them in the color map for colors that occur only once.

src/_color-themes.scss
JosefBud: @ayan4m1 Sorry I'm a little late here, but I have no problem with that at all!

src/components/ToastDrawer/ToastDrawer.scss
JosefBud: Can you give some examples of why it should be transparent? I just don't know why you'd need to be able to see through them when they're only up for a couple of seconds.

hylophile: What I meant is that they were transparent already before and I recreated this transparency explicitly, because I didn't want to change the current style. At least that's how it was displayed on my system.
I will remove this then, because I don't see a good reason for them being transparent either.

JosefBud: Sounds good, sorry for the confusion!

src/components/App/App.js
```diff
@@ -84,4 +94,4 @@ const mapDispatchToProps = dispatch => ({
   actions: bindActionCreators(appActions, dispatch)
 });
 
-export default withRouter(connect(null, mapDispatchToProps)(App));
+export default withTheme(withRouter(connect(null, mapDispatchToProps)(App)));
```
ayan4m1, line 97: 👌

src/components/Theme/Theme.js
```diff
@@ -0,0 +1,27 @@
...
+
+export const theme = WrappedComponent =>
+  class extends Component {
+    static displayName = 'Theme';
+    static propTypes = {
+      theme: PropTypes.string
+    };
+    render() {
+      return (
+        <BodyClassName className={'theme--' + this.props.theme}>
```
ayan4m1, line 15: I think now that we are using react-body-classname, you can just inline this functionality in App.js - there's a lot of boilerplate required both here and in this component's tests that could be dispensed with since the functionality is only ever used at the application level

src/reducers/theme.js
```diff
@@ -0,0 +1,34 @@
...
+};
+
+export const initialState = {
+  name: 'default'
+};
+
+export const reducer = (state = initialState, action = {}) => {
+  switch (action.type) {
+    case types.TOGGLE_DARK_MODE:
+      if (state.name === 'default') {
```
ayan4m1, line 20: Use a switch statement here instead of an `if { } else { }` block

src/reducers/theme.test.js
```diff
@@ -0,0 +1,17 @@
...
+  it('has TOGGLE_DARK_MODE action', () => {
+    expect(actions.toggleDarkMode()).toEqual({
+      type: types.TOGGLE_DARK_MODE
+    });
+  });
+
+  it('reduces TOGGLE_DARK_MODE action', () => {
+    const action = actions.toggleDarkMode();
+
+    expect(reducer({}, action)).toEqual({ name });
```
ayan4m1, line 15: It would improve this test if it also tested that the theme is set to `dark` if toggleDarkMode is called again.

src/components/App/App.test.js
```diff
@@ -44,10 +45,13 @@ describe('<App />', () => {
     initApp: jest.fn()
   };
   const mockStore = configureStore();
-  const store = mockStore({ application: initialState });
+  const store = mockStore({
+    application: initialApplicationState,
+    theme: initialThemeState
```
ayan4m1, line 50: Is there any specific reason for theme to live outside application?

src/components/ToastDrawer/ToastDrawer.scss
```diff
@@ -4,3 +4,24 @@
   z-index: 100;
   padding-right: 40px;
 }
+
+.toast-header {
```
ayan4m1, line 8: These two rules can be combined

src/pages/user/Settings.js
```diff
@@ -106,6 +117,21 @@ export default class UserSettings extends Component {
               </InputGroup>
             </Col>
           </Row>
+          <Row className="text-center">
+            <Col md="12" className="align-self-center">
+              Toggle Dark Mode
+              <ToggleButton
```
ayan4m1, line 123: 💯

src/pages/user/Settings.js
```diff
@@ -106,6 +117,21 @@ export default class UserSettings extends Component {
...
+                value={this.props.theme === 'default' ? false : true}
+                onClick={() => this.props.actions.toggleDarkMode()}
```
ayan4m1, line 125: Would you mind destructuring `this.props.actions` at the top of the render?

src/pages/user/Settings.test.js
```diff
@@ -0,0 +1,22 @@
...
+describe('<UserSettings />', () => {
+  const actions = {
+    toggleDarkMode: jest.fn()
+  };
+  const props = {
+    theme: 'default',
+    actions
+  };
+
+  const RoutedUserSettings = withMemoryRouter(UserSettings);
```
ayan4m1, line 15: Does this test give warnings/errors if you remove the `withMemoryRouter`?

src/components/App/App.test.js
hylophile: As we now don't have a theme component anymore, not really. I moved all theme-related redux code into application now, i.e. including the selectors/reducers/actions. The only issue is that the only place where the toggleDarkMode() action is used, on the user settings page, all application actions are imported. I don't know if that's an issue though (if it gets optimized away or if it would be better to only import one action or move the action to a different file).

src/pages/user/Settings.test.js
hylophile: Ah, nope. Removed!

src/pages/user/Settings.js
hylophile: Sure.

PR #74, opened 2020-04-23
Set max-width for Pager Elements

daviddyess:
Fixes an issue where Firefox ignores a width property for inline elements unless it sets a length value:

https://developer.mozilla.org/en-US/docs/Web/CSS/width

### DIFF
--- a/src/components/Pagination/Pagination.scss
+++ b/src/components/Pagination/Pagination.scss
@@ -1,12 +1,14 @@
 .pager {
   &-limit {
     display: inline;
-    width: 80px;
+    width: unset;
+    max-width: 80px;
   }
 
   &-page {
     display: inline;
-    width: 80px;
+    width: unset;
+    max-width: 80px;
     margin: 0 8px;
   }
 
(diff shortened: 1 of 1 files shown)


src/components/Pagination/Pagination.scss
```diff
@@ -1,12 +1,12 @@
 .pager {
   &-limit {
     display: inline;
-    width: unset;
+    width: 80px;
```
ayan4m1, line 4: Does this work in mobile?

daviddyess: FYI, I adjusted this a bit so this change wouldn't affect the browsers `unset` works in and added `max-width` to fix the issue in Firefox.

daviddyess: Toggling between `master` and the PR, the displays in a mobile view port size are the same for Chrome and the fix works in Firefox. I can do some adjustments if you want, as the offset in `xs` pushes everything to the right on smaller viewports.

PR #75, opened 2020-05-15
Flavor Stash / Notes

daviddyess:
User's Flavor Stash, with editors for min/max percentage and notes. Each flavor gets a single note, which utilizes markdown for rich text.


src/components/FlavorStash/Note.js
```diff
@@ -0,0 +1,230 @@
...
+    loading: PropTypes.any.isRequired,
+    note: PropTypes.object.isRequired,
+    userId: PropTypes.string.isRequired
+  };
+
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      editingNote: false,
```
ayan4m1, line 32: Could just be `editing`

src/components/FlavorStash/Note.js
```diff
@@ -0,0 +1,230 @@
...
+            onClick={() => this.handleNoteEditor(id)}
+          >
+            <FontAwesomeIcon icon="pen" size="sm" title="Edit Flavor Note" />
+          </Button>
+        )}
+      </Fragment>
+    );
+  }
+
+  handleNoteEditor() {
```
ayan4m1, line 81: Maybe `handleToggleEditing`?

src/components/FlavorStash/Note.js
```diff
@@ -0,0 +1,230 @@
...
+    const { note, update } = values;
+
+    if (!note) {
+      actions.deleteNote(values);
+    } else if (update === true) {
+      actions.updateNote(values);
+    } else {
+      actions.createNote(values);
+    }
+    this.setState({ updatedNote: note || '' });
```
ayan4m1, line 96: Why are we copying the note to local component state?

src/components/FlavorStash/Note.js
```diff
@@ -0,0 +1,230 @@
...
+                )}
+              </Field>
+            </Form.Row>
+            {values?.note ? (
+              <Form.Row>
+                <h4>Preview:</h4>
+                <Col md="12">
+                  {
+                    /* eslint-disable-next-line no-sync */
+                    unified()
```
ayan4m1, line 138: My two cents, this shouldn't be done synchronously or inline. Maybe use an async function that updates local component state with the parsed result?

src/components/FlavorStash/Note.js
```diff
@@ -0,0 +1,230 @@
...
+                    variant="primary"
+                    type="submit"
+                    disabled={submitting}
+                  >
+                    <FontAwesomeIcon icon="save" size="sm" />
+                    &nbsp;<span>Save</span>
+                  </Button>
+                  <Button
+                    onClick={e => this.handleNoteEditor(false, e)}
+                    className="button-animation button--cancel"
```
ayan4m1, line 169: This is likely something for a different pr, but I think this class name should be `button--animated`

src/components/FlavorStash/Note.js
```diff
@@ -0,0 +1,230 @@
...
+            {note[flavorId] || updatedNote
+              ? this.editNoteIcon(flavorId)
+              : this.addNoteIcon(flavorId)}{' '}
+            Flavor Notes
+          </h3>
+          {!editingNote && (note[flavorId] || updatedNote) ? (
+            <Fragment>
+              {
+                /* eslint-disable-next-line no-sync */
+                unified()
```
ayan4m1, line 200: Same thing here with the inline/sync parsing.

src/components/FlavorStash/Note.js
```diff
@@ -0,0 +1,230 @@
...
+
+export const mapStateToProps = state => ({
+  loading: isLoading(state),
+  note: getFlavorNote(state)
+});
+
+export const mapDispatchToProps = dispatch => ({
+  actions: bindActionCreators(
+    {
+      ...noteActions
```
ayan4m1, line 224: No need to spread if it's the only action group being used.

src/components/FlavorStash/Note.test.js
```diff
@@ -0,0 +1,135 @@
...
+    expect(actions.requestNote).toHaveBeenCalled();
+  });
+
+  it('can render with an empty note', () => {
+    const tree = renderer.create(<RoutedNote {...props} note={{}} />).toJSON();
+
+    expect(tree).toMatchSnapshot();
+  });
+
+  it('calls requestStash on mount', () => {
```
ayan4m1, line 53: Should be `requestNote` right?

src/pages/user/FlavorStash.js
```diff
@@ -1,63 +1,347 @@
...
 
     if (!stashLoaded) {
       actions.requestStash();
     }
   }
 
+  date(d) {
+    const s = new Date(d);
+
+    return s.toLocaleDateString();
```
ayan4m1, line 65: Use date-fns `format` for this

src/pages/user/FlavorStash.js
```diff
@@ -1,63 +1,347 @@
...
 
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      expanded: {},
+      editingStash: false,
+      removed: {},
+      usage: {},
+      viewingNote: {}
```
ayan4m1, line 44: This doesn't seem to be used, is it leftover?

src/pages/user/FlavorStash.js
```diff
@@ -1,63 +1,347 @@
...
+                {({ input }) => (
+                  <Form.Group as={Col}>
+                    <InputGroup>
+                      <InputGroup.Prepend>
+                        <InputGroup.Text id="inputGroupPrepend">
+                          Max
+                        </InputGroup.Text>
+                      </InputGroup.Prepend>
+                      <Form.Control {...input} type="number" step="0.1" />
+                      <InputGroup.Text id="inputGroupAppend">%</InputGroup.Text>
```
ayan4m1, line 202: Shouldn't need the `id=` attribute here.

src/pages/user/FlavorStash.js
```diff
@@ -1,63 +1,347 @@
...
+                    </InputGroup>
+                  </Form.Group>
+                )}
+              </Field>
+              <Field name="maxMillipercent">
+                {({ input }) => (
+                  <Form.Group as={Col}>
+                    <InputGroup>
+                      <InputGroup.Prepend>
+                        <InputGroup.Text id="inputGroupPrepend">
```
ayan4m1, line 197: Shouldn't need the `id=` attribute here or above.

src/components/FlavorStash/Note.js
daviddyess: It's mostly for user experience: it makes the update instant instead of having to pull it back from the API. But it also prevents the need for another GET request.

src/pages/user/FlavorStash.js
daviddyess: This seems to be something we should probably work on and integrate later. date-fns doesn't seem to be flexible enough to be used within the scope of this PR.

src/components/FlavorStash/Note.js
daviddyess: Also fixed in the page as well.

src/components/FlavorStash/Note.js
daviddyess: I've refactored this, when you create or update a note it updates the collection in the reducer

src/components/FlavorStash/Note.js
daviddyess: I haven't been able to get this to work with async. I also haven't been able to find an async example, they all show using this method

src/components/FlavorStash/Note.js
```diff
@@ -0,0 +1,183 @@
...
+                    /* eslint-disable-next-line no-sync */
+                    unified()
+                      .use(parse)
+                      .use(remark2react)
+                      .processSync(values.note).result
+                  }
+                </Col>
+              </Form.Row>
+            ) : null}
+            <Field name="userId">
```
ayan4m1, line 123: Can these fields be pulled at submit-time rather than populated as hidden fields now?

daviddyess: I removed them and added a commit, their values are pre-filled and submitted with the form. I mainly just had them there for readability in the generated page source.

PR #76, opened 2020-05-15
Added User Image Dropdown to Header

daviddyess:
Replaces the `User` dropdown menu on the Header component with a user image dropdown. If the user doesn't have a gravatar image, a generic user image displays, instead of the generic G icon often seen from gravatar. This change also moves the dropdown to the right side of the Header.


src/components/Header/Header.js
```diff
@@ -121,11 +151,12 @@ export class Header extends Component {
...
 
 const mapStateToProps = state => ({
-  loggedIn: isLoggedIn(state)
+  loggedIn: isLoggedIn(state),
+  user: getUser(state)
 });
 
 const mapDispatchToProps = dispatch => ({
-  actions: bindActionCreators(appActions, dispatch)
+  actions: bindActionCreators({ ...appActions }, dispatch)
```
ayan4m1, line 159: No need to spread this here

src/components/Header/Header.test.js
```diff
@@ -45,4 +46,19 @@ describe('<Header />', () => {
...
+      actions,
+      loggedIn: true,
+      user: { id: 1, emailAddress: 'jest@mixnjuice.com' }
+    };
+    const component = renderer.create(<RoutedHeader {...props} />);
+    const { instance } = component.root.findByType(Header);
+
+    expect(instance).toBeDefined();
+    expect(instance.gravatar).toEqual(
+      'https://www.gravatar.com/avatar/6114a57951ce2eefc79a1b3d7918a6f5?s=50&d=mp'
```
ayan4m1, line 61: 💯

PR #86, opened 2020-09-02
Vendor & Vendors Redux

daviddyess:
Adds Redux for flavor vendors

### DIFF
--- a/src/reducers/index.js
+++ b/src/reducers/index.js
@@ -14,6 +14,8 @@ import { reducer as recipe, initialState as recipeState } from './recipe';
 import { reducer as roles, initialState as rolesState } from './roles';
 import { reducer as toast, initialState as toastState } from './toast';
 import { reducer as users, initialState as usersState } from './users';
+import { reducer as vendor, initialState as vendorState } from './vendor';
+import { reducer as vendors, initialState as vendorsState } from './vendors';
 
 export const initialState = {
   application: applicationState,
@@ -24,7 +26,9 @@ export const initialState = {
   recipe: recipeState,
   roles: rolesState,
   toast: toastState,
-  users: usersState
+  users: usersState,
+  vendor: vendorState,
+  vendors: vendorsState
 };
 
 export default combineReducers({
@@ -36,5 +40,7 @@ export default combineReducers({
   recipe,
   roles,
   toast,
-  users
+  users,
+  vendor,
+  vendors
 });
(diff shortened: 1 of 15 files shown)


src/reducers/index.js
```diff
@@ -36,5 +40,7 @@ export default combineReducers({
   recipe,
   roles,
   toast,
-  users
+  users,
+  vendor,
+  vendors
```
ayan4m1, line 45: Is there a test that needs updating for this?

src/reducers/vendor.test.js
```diff
@@ -0,0 +1,102 @@
...
+  it('has DELETE_VENDOR action', () => {
+    expect(actions.deleteVendor({ vendorId, name })).toEqual({
+      type: types.DELETE_VENDOR,
+      vendorId,
+      name
+    });
+  });
+
+  it('reduces DELETE_VENDOR action', () => {
+    const action = actions.deleteVendor({});
```
ayan4m1, line 98: This should test that the vendor is actually removed from the collection

src/sagas/vendor.js
```diff
@@ -0,0 +1,177 @@
...
+      if (result.success) {
+        const {
+          response: { data }
+        } = result;
+
+        yield put(actions.requestVendorSuccess(data));
+      } else if (result.error) {
+        throw result.error;
+      } else {
+        throw new Error('Failed to get vendor!');
```
ayan4m1, line 28: The case where `vendor.vendorId === vendorId` probably shouldn't produce an error, but should be handled silently, since the data requested is already there... or maybe it should re-fetch it anyway, either way a little more thought is needed here.

src/sagas/vendor.js
```diff
@@ -0,0 +1,177 @@
...
+    }
+  } catch (error) {
+    const { message } = error;
+
+    yield put(actions.requestFailure(error));
+    yield call(helper.errorToast, message);
+  }
+}
+
+function* updateVendorWorker({ details: { vendorId, name, slug, code } }) {
```
ayan4m1, line 76: Should this be `{ vendor: { id, name, slug, code } }` ? Trying to standardize the object structures a bit so we don't have to do so much transformation.

src/sagas/vendors.test.js
```diff
@@ -0,0 +1,98 @@
...
+
+  const pager = {
+    count: 25,
+    limit: 20,
+    page: 1,
+    pages: 2
+  };
+
+  it('handles success in requestVendorsWorker', () => {
+    /* const gen = workers.requestVendorsWorker({ pager });
```
ayan4m1, line 18: Commented out code

src/reducers/index.js
daviddyess: Now that you mention it... it seems like there was a test for reducers' index, but it's not in the master branch.

PR #88, opened 2020-11-08
Use node LTS version instead of latest

ayan4m1:
This PR fixes CircleCI build failures related to using Node 15 which will be resolved if we build the app using Node LTS version instead.


daviddyess (approved):
:1st_place_medal:

PR #90, opened 2020-11-17
Class Comp -> FC -- IngredientList

rtragle:
This PR changes the IngredientList component from class based to functional w/ hooks

IngredientList FC refactor-2020-11-17_13 50 34

https://app.clickup.com/t/g6vu77


package-lock.json
```diff
@@ -3751,26 +3751,6 @@
         }
       }
     },
-    "@testing-library/user-event": {
```
ayan4m1, line 3754: Why is package-lock.json changing without a change to package.json?

ayan4m1 (approved):
Looks good, thanks! 🚀

package-lock.json
rtragle: I was wondering the same thing... Do you want me to remove this change from the PR?

ayan4m1 (approved):
Thanks for making those changes!

PR #92, opened 2020-11-19
Class Comp -> FC -- Page > Flavors

rtragle:
Refactor Flavors Page to FC

Flavors Page FC refactor-2020-11-18_17 00 47

Note: Pagination component will be refactored later.

Pork barrel:
1) snapshot-diff - added this jest extension to support `toMatchDiffSnapshot` to allow smaller snapshots
2) added data-testid to `<ToggleButton />`, so it's easier to query the ToggleButton for testing

https://app.clickup.com/t/g6vvmw

### DIFF
--- a/src/components/ToggleButton/ToggleButton.js
+++ b/src/components/ToggleButton/ToggleButton.js
@@ -30,7 +30,7 @@ const defaultProps = {
   iconProps: {},
   value: false,
   variant: 'check',
-  'data-testid': null
+  'data-testid': 'toggle-button'
 };
 
 function IconOnly(props) {
@@ -102,7 +102,7 @@ function ButtonElement(props) {
       title={title}
       onClick={onClick}
       className={classNames(className, 'btn-toggle')}
-      data-testid={props['data-testid'] ?? 'toggle-button'}
+      data-testid={props['data-testid']}
     >
       <FontAwesomeIcon size="lg" {...iconProps} icon={icon} />
     </Button>
--- a/src/components/ToggleButton/ToggleButton.js
+++ b/src/components/ToggleButton/ToggleButton.js
@@ -16,7 +16,8 @@ const propTypes = {
   value: PropTypes.bool,
   iconOnly: PropTypes.bool,
   variant: PropTypes.oneOf(toggleButtonProps.variants),
-  onClick: PropTypes.func
+  onClick: PropTypes.func,
+  'data-testid': PropTypes.string
 };
 
 const propTypesWithIcon = {
@@ -28,10 +29,13 @@ const defaultProps = {
   buttonProps: {},
   iconProps: {},
   value: false,
-  variant: 'check'
+  variant: 'check',
+  'data-testid': null
 };
 
-function IconOnly({ icon, iconProps, title, onClick, className }) {
+function IconOnly(props) {
+  const { icon, iconProps, title, onClick, className } = props;
+
   return (
     <FontAwesomeIcon
       size="lg"
@@ -40,6 +44,7 @@ function IconOnly({ icon, iconProps, title, onClick, className }) {
       title={title}
       onClick={onClick}
       className={classNames(className, 'icon-toggle')}
+      data-testid={props['data-testid']}
     />
   );
 }
@@ -52,7 +57,13 @@ function Switch(props) {
 
   return (
     <label className="switch my-auto mx-1">
-      <input type="checkbox" checked={value} onChange={onClick} name={title} />
+      <input
+        type="checkbox"
+        checked={value}
+        onChange={onClick}
+        name={title}
+        data-testid={props['data-testid']}
+      />
       <span className={classNames(className, 'slider slider--round')}></span>
     </label>
   );
@@ -88,10 +99,10 @@ function ButtonElement(props) {
   return (
     <Button
       {...buttonProps}
-      data-testid="toggle-button"
       title={title}
       onClick={onClick}
       className={classNames(className, 'btn-toggle')}
+      data-testid={props['data-testid'] ?? 'toggle-button'}
     >
       <FontAwesomeIcon size="lg" {...iconProps} icon={icon} />
     </Button>
--- a/src/components/ToggleButton/ToggleButton.js
+++ b/src/components/ToggleButton/ToggleButton.js
@@ -17,7 +17,7 @@ const propTypes = {
   iconOnly: PropTypes.bool,
   variant: PropTypes.oneOf(toggleButtonProps.variants),
   onClick: PropTypes.func,
-  'data-testid': PropTypes.string
+  testId: PropTypes.string
 };
 
 const propTypesWithIcon = {
@@ -30,12 +30,10 @@ const defaultProps = {
   iconProps: {},
   value: false,
   variant: 'check',
-  'data-testid': 'toggle-button'
+  testId: 'toggle-button'
 };
 
-function IconOnly(props) {
-  const { icon, iconProps, title, onClick, className } = props;
-
+function IconOnly({ icon, iconProps, title, onClick, className, testId }) {
   return (
     <FontAwesomeIcon
       size="lg"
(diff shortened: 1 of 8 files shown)


src/pages/Flavors.test.js
```diff
@@ -1,36 +1,33 @@
...
+    ...jest.requireActual('react-redux'),
+    useDispatch: jest.fn(() => dispatch),
+    useSelector: jest.fn()
+  };
+});
 
 jest.mock('components/Pagination/Pagination', () => {
-  const pagination = require('utils/testing').mockComponent('Pagination');
-  const PagerInfo = require('utils/testing').mockComponent('PagerInfo');
+  const pagination = () => <div>Pagination</div>;
```
rtragle, line 20: `require('utils/testing').mockComponent('Pagination')` doesn't work right w/ `@testing-library/react`

src/components/ToggleButton/ToggleButton.js
```diff
@@ -40,6 +44,7 @@ function IconOnly({ icon, iconProps, title, onClick, className }) {
       title={title}
       onClick={onClick}
       className={classNames(className, 'icon-toggle')}
+      data-testid={props['data-testid']}
```
ayan4m1, line 47: Maybe `testId` instead?

src/pages/Flavors.test.js
ayan4m1: You should be able to use `pagination` lowercase, but that's a good point.

src/pages/Flavors.test.js
```diff
@@ -1,36 +1,33 @@
...
+
+import PaginatedFlavors, { Flavors } from './Flavors';
+
+jest.mock('react-redux', () => {
+  const dispatch = jest.fn();
+
+  return {
+    ...jest.requireActual('react-redux'),
+    useDispatch: jest.fn(() => dispatch),
+    useSelector: jest.fn()
```
ayan4m1, line 15: Why do we need to mock useSelector here?

src/components/ToggleButton/ToggleButton.js
rtragle: I kind of like having the consistent `data-testid`. To answer your question, it needs to be passed in because I'm referencing it in unit tests

src/pages/Flavors.test.js
```diff
@@ -139,126 +125,98 @@ describe('Page <Flavors />', () => {
...
 
-    expect(tree).toMatchSnapshot();
+    expect(asFragment()).toMatchSnapshot();
   });
 
   it('can handle stashToggle', () => {
-    const event = {
-      target: { name: 'stashToggle', checked: true }
-    };
+    useSelector
```
rtragle, line 174: yeah, it'd be nice to have a good way to mock the useSelector

src/pages/Flavors.test.js
```diff
@@ -139,126 +152,92 @@ describe('Page <Flavors />', () => {
...
-    expect(
-      renderer
-        .create(
-          <Provider store={store}>
-            <RoutedConnectedFlavors />
-          </Provider>
-        )
-        .toJSON()
-    ).toMatchSnapshot();
+    expect(getByText(/log in/i)).toBeDefined();
```
rtragle, line 185: don't really like matching on text, but it seems best for this test case

src/pages/Flavors.js
```diff
@@ -1,198 +1,174 @@
...
+    if (loggedIn && !stashLoaded) {
+      dispatch(flavorActions.requestStash());
     }
-  }
-
-  handleStashToggle() {
-    const { stash } = this.props;
-    const { holdings: stashMap } = this.state;
-    const holdings = {};
+  }, [dispatch, loggedIn, stashLoaded]);
```
rtragle, line 64:
```
[dispatch, loggedIn, stashLoaded]
```
there's a performance issue with including `stashLoaded` in this list, but the eslint rules force me to do it. The behavior of the original component doesn't match this because the effect will be ran again every time stashLoaded set to false which happens on every stash update action in the reducer.

src/components/ToggleButton/ToggleButton.js
```diff
@@ -16,7 +16,8 @@ const propTypes = {
   value: PropTypes.bool,
   iconOnly: PropTypes.bool,
   variant: PropTypes.oneOf(toggleButtonProps.variants),
-  onClick: PropTypes.func
+  onClick: PropTypes.func,
+  'data-testid': PropTypes.string
```
ayan4m1, line 20: This should be `testId` or similar

src/components/ToggleButton/ToggleButton.js
ayan4m1: It's a React component prop name, it should not have a dash in it.

src/pages/Flavors.js
ayan4m1: Yeah, adding a comment and disabling the rule for that line sounds fine to me.

src/pages/Flavors.test.js
```diff
@@ -139,126 +152,92 @@ describe('Page <Flavors />', () => {
...
-    actions,
-    collection,
-    loggedIn: true,
-    pager,
-    pagerNavigation,
-    stash,
-    stashLoaded: false
-  };
+  const ReduxConnectedFlavors = (flavorsProps) => (
+    <Provider store={store}>
```
ayan4m1, line 166: We have a withProvider() utility in src/utils/testing.js that provides a way to do this

src/pages/Flavors.test.js
```diff
@@ -139,126 +152,92 @@ describe('Page <Flavors />', () => {
...
+      <Flavors {...flavorsProps} />
+    </Provider>
+  );
+
+  beforeEach(() => {
+    useDispatch().mockClear();
```
ayan4m1, line 172: Maybe jest.clearAllMocks() here?

src/pages/Flavors.test.js
ayan4m1, line 185: I think this should be `.toBeInTheDocument()`, no?

src/pages/Flavors.test.js
```diff
@@ -139,126 +152,92 @@ describe('Page <Flavors />', () => {
...
-    expect(
-      renderer.create(<RoutedFlavors {...notLoggedInProps} />).toJSON()
-    ).toMatchSnapshot();
+    const { asFragment } = render(<ReduxConnectedFlavors {...props} />);
+
+    expect(asFragment()).toMatchSnapshot();
   });
 
-  const component = renderer.create(<RoutedFlavors {...props} />);
+  it("matches snapshot diff of user's stash (after toggle)", () => {
```
ayan4m1, line 198: I don't understand what this test does based on this name. I think it "ensures toggle button click updates state"?

ayan4m1 (changes requested):
Looking pretty good, I have a couple questions and a couple notes for you.

src/pages/Flavors.test.js
rtragle: It toggles the button, and takes a snapshot of the difference between the html before and after the toggle

src/pages/Flavors.test.js
rtragle: I think you're right. I'll mess with it.

src/pages/Flavors.test.js
rtragle: good catch

src/components/ToggleButton/ToggleButton.js
rtragle: will fix

src/pages/Flavors.test.js
rtragle: yeah, you're right. I was trying that earlier, but I think I was using `toBeInDocument` w/o `The`, so it wasn't working right.. thanks for catching this

ayan4m1 (approved):
Great, thanks!

